PR #3, opened 2023-12-29
Implement versioning.

IMB11:


.vitepress/config.mts
```diff
@@ -23,17 +32,20 @@ export default defineConfig({
...
+    sidebar: {
+      '/': [
+        {
+          text: 'Examples',
+          items: [
+            { text: 'Develop Example', link: '/develop/example' }
+          ]
+        },
+      ]
+    },
```
modmuss50: In the example repo this had `generateVersionSidebars()` and the `generateVersionRewrites()`.

The sidebar was also defined in its own file, I suggested this could just be a sidebar.json file maybe even in the root of the repo?

IMB11: I turned the versioning system into a library as other vitepress users wanted to use it.

https://www.npmjs.com/package/vitepress-versioning-plugin

Versioned sidebars are in `.vitepress/sidebars/versioned/{version}.json`

I'll make a system to load sidebars from `~/sidebars.json`

modmuss50 (approved):

PR #7, opened 2023-12-29
Add Reference mod

modmuss50:
Currently just the example mod with the following changes:
- a 1.20.4 subproject, should this be called latest?
- Checkstyle, editorconfig, spotless.
- Datagen enabled.


IMB11:
(PS. might be worth adding @FabricMC/documentation to codeowners too?)

modmuss50:
> (PS. might be worth adding @FabricMC/documentation to codeowners too?)

Shouldnt be required I dont think.

IMB11 (approved):

PR #16, opened 2024-01-04
Contribution Guidelines

IMB11:


.vitepress/config.mts
```diff
@@ -46,7 +51,8 @@ export default defineVersionedConfig(__dirname, {
     // https://vitepress.dev/reference/default-theme-config
     nav: [
       { text: 'Home', link: 'https://fabricmc.net/' },
-      { text: 'Download', link: 'https://fabricmc.net/use' }
+      { text: 'Download', link: 'https://fabricmc.net/use' },
+      { text: 'Contribute', link: '/contributing' },
```
modmuss50: Nit: Likely fine for now, but might want to move this to a footer later on? Or we could maybe expand this to cover contributing to other projects. Just somehting to think about.

modmuss50 (dismissed):

.vitepress/config.mts
IMB11: Yeah, actually, could probably make this a dropdown

modmuss50 (approved):

PR #22, opened 2024-01-22
Events Guide

dicedpixels:
Ports the "Events" related pages from the Wiki. I did not include the [Event Index](https://fabricmc.net/wiki/tutorial:event_index), as it's draft itself and looks very painful to maintain.

Overall, very little changed. Minor code changes to support MC 1.20.x.

### DIFF
--- a/develop/events.md
+++ b/develop/events.md
@@ -1,0 +1,149 @@
+---
+title: Events
+description: A guide for using events provided by the Fabric API.
+authors:
+  - dicedpixels
+---
+
+# Events
+
+Fabric API provides a system that allows mods to react to actions or occurrences, also defined as *events* that occur in the game.
+
+Events are hooks that satisfy common use cases and/or provide enhanced compatibility and performance between mods that hook into the same areas of the code. The use of events often substitutes the use of mixins.
+
+Fabric API provides events for important areas in the Minecraft codebase that multiple modders may be interested in hooking into.
+
+Events are represented by instances of `net.fabricmc.fabric.api.event.Event` which stores and calls *callbacks*. Often there is a single event instance for a callback, which is stored in a static field `EVENT` of the callback interface, but there are other patterns as well. For example, `ClientTickEvents` groups several related events together.
+
+## Callbacks
+
+Callbacks are a piece of code that is passed as an argument to an event. When the event is triggered by the game, the passed piece of code will be executed.
+
+### Callback Interfaces
+
+Each event has a corresponding callback interface, conventionally named `<EventName>Callback`. Callbacks are registered by calling `register()` method on an event instance, with an instance of the callback interface as the argument. 
+
+All event callback interfaces provided by Fabric API can be found in the `net.fabricmc.fabric.api.event` package.
+
+## Listening to Events
+
+### A Simple Example
+
+This example registers an `AttackBlockCallback` to damage the player when they hit blocks that don’t drop an item when hand-mined.
+
+@[code lang=java transcludeWith=:::1](@/reference/latest/src/main/java/com/example/docs/event/Events.java)
+
+### Adding Items to Existing Loot Tables
+
+Sometimes you may want to add items to loot tables. For example, adding your drops to a vanilla block or entity.
+
+The simplest solution, replacing the loot table file, can break other mods. What if they want to change them as well? We’ll take a look at how you can add items to loot tables without overriding the table.
+
+We’ll be adding eggs to the coal ore loot table.
+
+#### Listening to Loot Table Loading
+
+Fabric API has an event that is fired when loot tables are loaded, `LootTableEvents.MODIFY`. You can register a callback for it in your mod initializer. Let’s also check that the current loot table is the coal ore loot table.
+
+```java
+LootTableEvents.MODIFY.register((resourceManager, lootManager, id, tableBuilder, source) -> {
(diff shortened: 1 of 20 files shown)


IMB11:
I feel like the EventIndex can be automated, or we can just link to the fabric-api javadoc instead - dont worry much about it.

develop/events.md
```diff
@@ -0,0 +1,160 @@
...
+The simplest solution, replacing the loot table file, can break other mods. What if they want to change them as well? We’ll take a look at how you can add items to loot tables without overriding the table.
+
+We’ll be adding eggs to the coal ore loot table.
+
+#### Listening to Loot Table Loading
+
+Fabric API has an event that is fired when loot tables are loaded, `LootTableEvents.MODIFY`. You can register a callback for it in your mod initializer. Let’s also check that the current loot table is the coal ore loot table.
+
+```java
+LootTableEvents.MODIFY.register((resourceManager, lootManager, id, tableBuilder, source) -> {
```
modmuss50: Should be backed by real code.

develop/events.md
```diff
@@ -0,0 +1,160 @@
...
+    }
+});
+```
+
+#### Adding Items to the Loot Table
+
+In loot tables, items are stored in *loot pool entries*, and entries are stored in *loot pools*. To add an item, we’ll need to add a pool with an item entry to the loot table. We can make a pool with `LootPool#builder`, and add it to the loot table.
+
+```java
+LootTableEvents.MODIFY.register((resourceManager, lootManager, id, tableBuilder, source) -> {
```
modmuss50: Same again.

reference/latest/src/main/java/com/example/docs/event/Events.java
```diff
@@ -0,0 +1,58 @@
...
+
+import net.fabricmc.fabric.api.event.player.AttackBlockCallback;
+import net.fabricmc.fabric.api.loot.v2.LootTableEvents;
+
+// Class to contain all mod events.
+public class Events {
+	private static final Identifier COAL_ORE_LOOT_TABLE_ID = Blocks.COAL_ORE.getLootTableId();
+
+	// Call this method in your initializer.
+	public static void registerModEvents() {
```
modmuss50: Would it be better to make this class a ModInitializer?

develop/events.md
```diff
@@ -0,0 +1,160 @@
...
+  - solidblock
+  - draylar
+  - jamieswhiteshirt
+  - PhoenixVX
+  - Juuxel
+  - YanisBft
+  - liach
+---
+
+<!-- Couldn't find GitHub usernames for: fudge, stormyfabric -->
```
IMB11: No worries, it seems StormyFabric has deleted their GitHub account. Fudge might be @natanfudge - waiting for confirmation.

develop/events.md
IMB11: Please place this in the reference mod.

develop/events.md
```diff
@@ -0,0 +1,160 @@
...
+    }
+}
+ 
+return ActionResult.PASS;
+```
+
+We can Javadoc comments to the top of callback classes to document what each `ActionResult` does. In our case, it might be:
+
+```java
+/**
```
IMB11: Use a code snippet here, this comment already exists in the `SheepShearCallback` class.

develop/events.md
```diff
@@ -0,0 +1,160 @@
...
+
+* Creating the event callback interface
+* Triggering the event from a mixin
+* Creating a test implementation
+
+### Creating the Event Callback Interface
+
+The callback interface describes what must be implemented by event listeners that will listen to your event. The callback interface also describes how the event will be called from our mixin. It is conventional to place an `Event` object as a field in the callback interface, which will identify our actual event.
+
+For our Event implementation, we will choose to use an array-backed event. The array will contain all event listeners that are listening to the event.
```
IMB11:
```suggestion
For our `Event` implementation, we will choose to use an array-backed event. The array will contain all event listeners that are listening to the event. 
```

IMB11 (changes requested):
Minor nitpicks. See modmuss's review for the major stuff.

develop/events.md
IMB11: Confirmed - `fudge` on the wiki is infact natanfudge.

reference/latest/src/main/java/com/example/docs/event/Events.java
IMB11: Probably, yeah. Might be worth compartmentalizing pages into seperate initializers for future reference.

develop/events.md
dicedpixels: My intention (rather the original Wiki's) was show the signature. Of course I can move this to the reference mod, but that would mean there's a random empty event in the code. Personally, I'd rather have the final example, which is linked from the reference mod.

reference/latest/src/main/java/com/example/docs/event/Events.java
dicedpixels: I forgot you can have multiple initializer. Yes, I'll move it.

develop/events.md
dicedpixels: I should have looked more into the code snippet plugin. It can do a bit more. So disregard my previous comment. All the code is now embedded from the reference mod.

IMB11 (changes requested):
Looks good! You just need to solve the checkstyle issues :+1:

IMB11 (approved):

PR #37, opened 2024-02-11
Develop Sidebar + Setup spanish for translators

IMB11:


.vitepress/sidebars/es/players.ts
```diff
@@ -0,0 +1,63 @@
+import { DefaultTheme } from "vitepress";
+
+export default [
```
modmuss50, line 3: This is fine for now, we can maybe look into a better way of translating the sidebar once crowdin is setup.

modmuss50 (approved):

PR #38, opened 2024-02-11
Warning Banner

IMB11:
Publishing script should work as is, provided by vitepress docs - just need to provide CNAME file via repo settings to display on docs.fabricmc.net


.github/workflows/publish.yml
```diff
@@ -0,0 +1,63 @@
+# Sample workflow for building and deploying a VitePress site to GitHub Pages
```
modmuss50: Can I do this please, just a little pedantic about making sure I understand what its doing. And if I make a mistake and it fucks something up its then my fault and not yours 👍

.vitepress/theme/components/BannerComponent.vue
```diff
@@ -0,0 +1,113 @@
...
+      '--vp-layout-top-height',
+      `${height.value + 16}px`
+    );
+  }
+});
+
+const dismiss = () => {
+  localStorage.setItem(
+    'dismissBanner',
+    (Date.now() + 8.64e7 * 1).toString() // current time + 1 day
```
modmuss50: This doesnt seem to be working.

why `* 1`?

`8.64e7` is not clear, just keep it simple and use something like: `24 * 60 * 60 * 1000`

.vitepress/theme/components/BannerComponent.vue
```diff
@@ -0,0 +1,113 @@
...
+  );
+  document.documentElement.classList.add('banner-dismissed');
+};
+
+onBeforeMount(() => {
+  if (localStorage.getItem('dismissBanner') !== null) {
+    var date = parseInt(localStorage.getItem('dismissBanner')!);
```
modmuss50: Nit, store the local stotrage item in a variable, there is a possible (unlikely) race condition here.

.vitepress/theme/components/BannerComponent.vue
```diff
@@ -0,0 +1,113 @@
...
+    if (Date.now() > date) {
+      dismiss();
```
modmuss50: This will set the date forward again, is this expected?

IMB11:
I've completely brain farted on this date logic, one moment.

modmuss50:
It still seems to be faslhing in for a moment, if its not easy to fix its not the end of the world, but it is a little jarring.

modmuss50 (approved):

PR #44, opened 2024-02-12
Dynamic locales + sidebar translations.

IMB11:
Sidebar translations exist in `translated/locale/sidebar_translations.json`

I have modified the crowdin file.

If you're confused about the new glob pattern for the translated folder ignore - that makes two of us. It just works :tm:

image


crowdin.yml
```diff
@@ -1,8 +1,8 @@
 files:
-  - source: /**/*.md
+  - source: /**/@(*.md|sidebar_translations.json)
```
modmuss50, line 2: I guess this works, it might have been cleaner to just have a seperate `-source` for the sidebar json?

translated/en_us/sidebar_translations.json
```diff
@@ -0,0 +1,34 @@
+{
```
modmuss50: It doesnt make sense for this file to be in /translated. As far as im concerned that entire directory shouldnt be touched by humans. I would maybe move this into the root, or alongside the sidebar ts files.

package.json
```diff
@@ -11,6 +11,7 @@
     "@rollup/rollup-linux-x64-gnu": "4.6.1"
   },
   "dependencies": {
+    "inter": "^2.1.0",
```
modmuss50, line 14: Is it possible to use: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames? Instead of pulling a 80MB library?

```js
const displayNames = new Intl.DisplayNames(['es'], { type: 'region' });
const displayName = displayNames.of('ES');
console.log(displayName);
```

.vitepress/i18n.ts
```diff
@@ -0,0 +1,89 @@
...
+    localeName = localeName.charAt(0).toUpperCase() + localeName.slice(1);
+
+    locales[folder] = {
+      label: localeName,
+      link: `/${folder}/`,
+      lang: folder,
+    }
+  }
+
+  console.log(locales);
```
modmuss50: Nit: remove

package.json
IMB11: It doesn't support the 4 letter locale codes. Intl only supports the old 2 letter locales, and only has locale names in english, making it useless for dynamic locales.

modmuss50 (approved):
Fine for now, I do want to look to remove the inter library as its massive.

PR #50, opened 2024-02-17
Damage Types Guide

dicedpixels:
Ports [Adding a Damage Type](https://fabricmc.net/wiki/tutorial:damagetypes).


reference/latest/src/main/resources/data/minecraft/tags/damage_type/bypasses_armor.json
```diff
@@ -0,0 +1,6 @@
+{
```
modmuss50: Everything should ideally be data generated. Even if its not documented.

dicedpixels: Tags can be easily done. Do you think damage_types should be data genned as well?

IMB11: Yes, I believe you can just use a tag provider that uses the damage type type. Correct me if im wrong!

IMB11: Edited comment.

dicedpixels: Everything is datagenned now. The doc itself will not mention anything about datagen to avoid complexity. Lets add a info box saying this too can be datagenned when the datagen docs are done.

IMB11 (approved):

develop/entities/damage-types.md
```diff
@@ -0,0 +1,111 @@
...
+The `RegistryKey` can be obtained as follows:
+
+@[code lang=java transcludeWith=:::1](@/reference/latest/src/main/java/com/example/docs/damage/FabricDocsReferenceDamageTypes.java)
+
+### Using Damage Types
+
+To demonstrate the use of custom damage types, we will use a custom block called _Tater Block_. Let's make is so that
+when a living entity steps on a _Tater Block_, it deals _Tater_ damage.
+
+::: info
```
IMB11: Remove this callout - block guide doesn't exist yet.

IMB11 (changes requested):

IMB11 (approved):

IMB11:
@modmuss50 LGTM

modmuss50 (approved):

PR #51, opened 2024-02-18
Decrease the tab size for code blocks

modmuss50:
I was unable to do this without using `!important`, if you can please let me know :D


IMB11:
Using `!important` is fine in this scenario as we're overriding the default vitepress styling.

modmuss50:
Closing this PR in favour of reducing the amount of indentation, 2 spaces can be hard to read.

PR #61, opened 2024-02-22
Creating Custom Particles Guide Overhaul

Superkat32:
I've updated the Creating Custom Particles guide to include multiple new things which I feel benefit the reader. These include:

- A better example particle, a blue sparkle. I've been reading some other guides on the Docs and really liked how the other guides used fun or somewhat helpful examples instead of something generic, like "my_particle". So I changed the example to be called the "sparkle particle" which is a shiny blue sparkle(similar to the green sparkles seen when bone-mealing anything). The hope is that this does a better job of getting creativity going for the reader.
- Easier to follow JSON and textures section. 3-4 pretty long sentences talking about files and folders wasn't my favorite thing, so I tried my best to make it better. Thanks IMB for the table suggestion.
- Particle showcase image.
- There's actually a quick explanation of what in the world a "particle factory" is now. This should hopefully help readers not feel confused about _why_ they're using the end rod particle's "particle factory".
- Tips and information boxes. They felt pretty fun and helpful at getting information across quickly.

I'd like for whoever confirms this to quickly look over everything. Specifically, the particle factory implementation tip. I really like it because I feel like it is a really helpful tip, but I also don't know if mentioning a) the hotkeys and or b) the tip itself is really needed. I decided I would get some feedback on that.


IMB11 (approved):
This is really good! It's a good improvement and it really improves the readability.

PR #77, opened 2024-03-25
Fix Codeowners file.

IMB11:
Allows people to edit sidebar without needing developer review.


modmuss50 (changes requested):
Can you just add the sidebar file/folders instead of defining everything else.

IMB11:
~~You can't, you'd still require the review from Developers as they have the whole .vitepress folder - also, the `!` negate syntax doesn't work with codeowners files.~~

Turns out exclusion syntax is a bit odd, but it's there.

PR #78, opened 2024-03-25
Exclude regional variants of languages on crowdin.

IMB11:


modmuss50 (approved):
Not idela, but bettter than nothing 👍

PR #83, opened 2024-04-04
Add robots.txt

IMB11:


public/robots.txt
```diff
@@ -0,0 +1,5 @@
+User-agent: *
+Allow: /$
+Disallow: /
```
modmuss50: What are you aiming to solve here? Im just reading and it says: 
> If you use them, be aware that most bots will not know how to deal with them. Most bots will ignore the Allow directives and treat $ as a literal.

I think we should make sure to keep this simple, and use tried and tested.

Can we not just have:
```
User-agent: *
Allow: /

Sitemap: https://docs.fabricmc.net/sitemap.xml 
```

IMB11: This will make all versioned pages be indexed alongside the sitemap.

By disallowing all pages except the sitemap, you exclude versioned pages.

The /$ only allows the root page. as Disallow: / excludes the root page, even if it's in the sitemap.

IMB11: We could just do Disallow: / and see what happens - the root page is in the sitemap, but im getting varying results, some bots are indexing it because it's in the sitemap, but some are ignoring it because it's Disallow: /

modmuss50: I worry that its not supported by all search engines, it seems to be google only. What would it take to add `<meta name="robots" content="noindex">` to all of the versioned pages?

IMB11: I could look into that yeah

modmuss50 (dismissed):
Happy to merge this now as we dont have other versioned pages yet.

IMB11:
I can add the meta noindex real quick - 1 moment.

IMB11:
Should be good now.

modmuss50 (approved):
Great, I didnt realise how easy that would be 👍

PR #99, opened 2024-05-11
Lint Markdown files, fix headings style

its-miroma:
## Linting

- Remove trailing whitespace
- Single newline at the end of each file
- Remove unnecessary quotes in frontmatters
- Use relative link for [CONTRIBUTING.md]
- Update [sidebar_translations.json]
- Increment headings one level at a time
- Remove punctuation from headings, except exclamation marks

## Headings

- Use Title Case through <https://www.titlecase.com/>
- Questions are exempted

## Other

- Capitalize various instances of "Java", "Loader", "Loom"
- Add custom heading link in [codecs.md]
- Avoid using emphasis for headings in [effects.md]
  - Side Effect: the auto-generated TOC now shows those two headings
- Change info box in [potions.md] (may fix #89)
- Add alt text to image in [hud.md]
- Avoid using invisible spaces for a line break in [creating-particles.md]
- Capitalize alt texts in [custom.md]

[CONTRIBUTING.md]: </CONTRIBUTING.md>
[sidebar_translations.json]: </sidebar_translations.json>
[codecs.md]: </develop/codecs.md#mapcodec>
[effects.md]: </develop/entities/effects.md>
[potions.md]: </develop/items/potions.md>
[hud.md]: </develop/rendering/hud.md>
[creating-particles.md]: </develop/rendering/particles/creating-particles.md>
[custom.md]: </develop/sounds/custom.md>

PR #101, opened 2024-05-12
Add GitHub workflow for `markdownlint`

its-miroma:


.github/workflows/markdownlint.yml
```diff
@@ -0,0 +1,51 @@
...
+      - '!versions/**'
+  workflow_dispatch: # Manual run
+
+concurrency:
+  group: markdownlint-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  markdownlint:
+    runs-on: ubuntu-latest
```
modmuss50: Specify version

.github/workflows/markdownlint.yml
```diff
@@ -0,0 +1,51 @@
...
+      - id: node
+        if: steps.changes.outputs.any_changed == 'true'
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: .nvmrc
+        continue-on-error: true  # lint even without `titlecase`
+
+      - id: titlecase
+        if: steps.node.outcome == 'success'
+        run: npm install markdownlint-rule-titlecase
```
modmuss50: Specify version

.github/workflows/markdownlint.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - id: lint
+        if: steps.changes.outputs.any_changed == 'true'
+        uses: DavidAnson/markdownlint-cli2-action@v16
+        with:
+          config: .markdownlint-cli2.yaml
+          globs: ${{ steps.changes.outputs.all_changed_files }}
```
modmuss50: Is there a good reason to only lint changed files? It would be easier to just do it all, and would also make any changes to the lint rules also apply to existing files.

its-miroma: Performance reasons mainly, and also if all files are linted contributors may have to fix issues in unrelated files

A pull request shouldn't get blocked by this action because an unrelated file is not linted

its-miroma: I guess that wouldn't be a problem if all files were linted to begin with, though

its-miroma: > would also make any changes to the lint rules also apply to existing files

See #99, there should not be such problem after that's merged

IMB11 (approved):

IMB11:
LGTM, #99 has been merged.

.github/workflows/markdownlint.yml
```diff
@@ -0,0 +1,51 @@
...
+      - id: changes
+        uses: tj-actions/changed-files@v44
+        with:
+          files: |
+            **/*.md
+            !translated/**
+            !versions/**
+
+      - id: node
+        if: steps.changes.outputs.any_changed == 'true'
```
modmuss50: I would just always lint the entire thing, we arent too bothered about speed here. Keep it simple.

.github/workflows/markdownlint.yml
```diff
@@ -0,0 +1,51 @@
...
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: .nvmrc
+        continue-on-error: true  # lint even without `titlecase`
+
+      - id: titlecase
+        if: steps.node.outcome == 'success'
+        run: npm install markdownlint-rule-titlecase@v0.1.0
```
modmuss50, line 33: What does this do? is this magically picked up by the next step?

modmuss50 (changes requested):

.github/workflows/markdownlint.yml
IMB11: Some translations will be unlintable based on the way the translations are patched in by Crowdin, not nessecarily about speed. I agree that we should lint versioned pages if they're edited (just not translation pages from Crowdin)

IMB11 (changes requested):
See comment by modmuss on linting everything.

.github/workflows/markdownlint.yml
its-miroma: To make sure that all headings use Title Case, I used a custom `markdownlint` rule.

> is this magically picked up by the next step?

See <https://github.com/its-miroma/fabric-docs/blob/e1c496ae37812f77bfa68cba9b4419af352c3af2/.markdownlint-cli2.yaml#L50-L51>

.github/workflows/markdownlint.yml
its-miroma: Does linting (aka modifying) translated files cause problems in Crowdin? That's my only concern

its-miroma: > I would just always lint the entire thing

Even unchanged files? See <https://github.com/FabricMC/fabric-docs/pull/101#discussion_r1597707214>

IMB11: > Does linting (aka modifying) translated files cause problems in Crowdin? That's my only concern

No, its more the other way round, the linter would error out pretty much instantly with most of the non-latin languages (chinese, etc.), I had a try and it just doesn't support them very well.

IMB11: > Even unchanged files? See https://github.com/FabricMC/fabric-docs/pull/101#discussion_r1597707214

A good point was brought up though, if we change lint rules, it'd enforce us to change the old pages :+1:

IMB11: tldr:

- Lint all files.
- Ignore anything in versions/<version>/translated/
- Ignore anything in translated/

IMB11 (dismissed):

IMB11:
This seems good, LGTM.

modmuss50 (approved):

PR #107, opened 2024-05-29
Use relative links

its-miroma:
Fix some links.


IMB11 (changes requested):
Use ./develop/index.md, the relative links you've specified wont work.

IMB11 (approved):

PR #109, opened 2024-06-12
Added Mixin tutorial + 1 more

ghost:
- Fixed mod crash due to registering sounds twice
- Stopped lerp-rgb square in hud showing when running the mod, code is still present and shows exactly the same on the hud tutorial page

- Mixin tutorial, as follows:
Mixin tutorial

### DIFF
--- a/develop/sounds/custom.md
+++ b/develop/sounds/custom.md
@@ -60,4 +60,4 @@ This way, the `ModInitializer` implementing entrypoint class needs to only imple
 
 ## Using the Custom SoundEvent
 
-Use the helper class to access the custom SoundEvent. Checkout the [Playing SoundEvents](/develop/sounds/using-sounds) page to learn how to play sounds.
+Use the helper class to access the custom SoundEvent. Checkout the [Playing SoundEvents](./using-sounds.md) page to learn how to play sounds.
--- a/develop/sounds/using-sounds.md
+++ b/develop/sounds/using-sounds.md
@@ -21,7 +21,7 @@ The `playSound()` method is used with the `LivingEntity` object. Only the SoundE
 
 ### SoundEvent and SoundCategory
 
-The SoundEvent defines which sound will be played. You can also [register your own SoundEvents](./custom) to include your own sound.
+The SoundEvent defines which sound will be played. You can also [register your own SoundEvents](./custom.md) to include your own sound.
 
 Minecraft has several audio sliders in the in-game settings. The `SoundCategory` enum is used to determine which slider will adjust your sound's volume.
 
--- a/index.md
+++ b/index.md
@@ -14,12 +14,12 @@ features:
   - title: Developer Guides
     icon: 🛠️
     details: Our curated developer guides, written by the community, span a wide range of topics from setting up a development environment to more advanced topics, such as rendering and networking.
-    link: /develop/
+    link: ./develop/index.md
     linkText: Get Started
   - title: Player Guides
     icon: 📚
     details: Are you a player looking to use mods powered by Fabric? Our player guides have you covered. These guides will help you in downloading, installing, and troubleshooting Fabric mods.
-    link: /players/
+    link: ./players/index.md
     linkText: Read More
 ---
 


develop/using-mixins.md
```diff
@@ -0,0 +1,53 @@
+---
+title: Using Mixins
```
Octol1ttle: Having the title just `Mixins` may be more preferable, as titles of other pages in this section don't use verbs

develop/using-mixins.md
```diff
@@ -0,0 +1,53 @@
...
+  ...,
+  "mixins": [
+    "ArrowEntityMixin"
+  ],
+  ...
+}
+```
+:::
+
+At the moment, we've targetted the class but haven't actually made it do anything. For this we need a method annotated with `@Inject`, as follows.
```
Octol1ttle:
```suggestion
At the moment, we've targeted the class but haven't actually made it do anything. For this we need a method annotated with `@Inject`, as follows.
```

Octol1ttle: > For this we need a method annotated with `@Inject`, as follows.

This sentence needs to emphasize that `@Inject` is used here *for the purpose of this example,* and that other annotations are used where applicable

develop/using-mixins.md
```diff
@@ -0,0 +1,53 @@
...
+
+@[code lang=java transcludeWith=:::1](@/reference/latest/src/main/java/com/example/docs/mixin/ArrowEntityMixin.java)
+
+This may seem complicated at first, but it is relatively simple once you break it down into its components.
+- `at = @At("TAIL")`: Every injection requires an injection point, possible values include, but are not limited to `HEAD`, `TAIL`, `INVOKE` and `RETURN`. `TAIL` selects the end of the method, regardless of where it returns. In this case, returns do not matter, since the target method never calls `return;`.
```
Octol1ttle:
```suggestion
- `at = @At("TAIL")`: Every injection requires an injection point, possible values include, but are not limited to `HEAD`, `TAIL`, `INVOKE` and `RETURN`. `TAIL` selects the end of the method, regardless of where it returns. In this case, returns do not matter, since the target method never uses `return;`.
```

develop/using-mixins.md
```diff
@@ -0,0 +1,53 @@
...
+- `method = "onHit"`: Since the method is called `onHit`, and we want to inject into it, we write `onHit` in the method parameter. As it is the only method called `onHit`, it isn't necessary to specify its parameters, but when it is, the plugins' suggestions will help you a lot.
+- `LivingEntity target, CallbackInfo ci`: Every inject-annotated method is passed a `CallbackInfo` object, which can be used to return methods when the injected method has the `cancellable = true` parameter. It also passes the target method's parameter, a `LivingEntity`, which is used to summon the explosion.
```
Octol1ttle:
```suggestion
- `LivingEntity target, CallbackInfo ci`: Every inject-annotated method is passed a `CallbackInfo` object, which can be used to return from methods when the injection has the `cancellable = true` parameter. It also passes the target method's parameter, a `LivingEntity`, which is used to summon the explosion.
```

develop/using-mixins.md
```diff
@@ -0,0 +1,53 @@
...
+- `target.getY() + (target.getHeight() / 2)`: Entity y position is below the feet, by adding half the entity's height, the explosion is summoned in the middle of the entity.
```
Octol1ttle:
```suggestion
- `target.getY() + (target.getHeight() / 2)`: Entities' Y position is located at their feet level. We make the explosion appear in the middle of the entity by adding half the entity's height.
```

develop/using-mixins.md
```diff
@@ -0,0 +1,53 @@
...
+
+As we can see by testing the mod, arrows now detonate an explosion upon hitting an entity.
```
Octol1ttle:
```suggestion
As we can see by testing the mod, arrows now detonate upon hitting an entity.
```
or
```suggestion
As we can see by testing the mod, arrows now summon an explosion upon hitting an entity.
```

reference/latest/src/main/java/com/example/docs/mixin/ArrowEntityMixin.java
```diff
@@ -0,0 +1,22 @@
...
+import net.minecraft.entity.projectile.ArrowEntity;
+import net.minecraft.world.World;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+// :::1
+@Mixin(ArrowEntity.class)
+public class ArrowEntityMixin {
```
Octol1ttle:
```suggestion
public abstract class ArrowEntityMixin {
```

reference/latest/src/main/java/com/example/docs/mixin/ArrowEntityMixin.java
```diff
@@ -0,0 +1,22 @@
...
+
+	@Inject(at = @At("TAIL"), method = "onHit")
+	private void summonExplosion(LivingEntity target, CallbackInfo ci) {
```
Octol1ttle:
```suggestion
	private void explodeOnHit(LivingEntity target, CallbackInfo ci) {
```
or
```suggestion
	private void summonExplosionOnHit(LivingEntity target, CallbackInfo ci) {
```

reference/latest/src/main/java/com/example/docs/mixin/ArrowEntityMixin.java
```diff
@@ -0,0 +1,22 @@
...
+		World world = target.getWorld();
+		world.createExplosion(target, target.getX(), target.getY() + (target.getHeight() / 2), target.getZ(), 5.0F, World.ExplosionSourceType.NONE);
+	}
+
```
Octol1ttle:
```suggestion
```

reference/latest/src/main/java/com/example/docs/mixin/ArrowEntityMixin.java
```diff
@@ -0,0 +1,22 @@
...
+import net.minecraft.world.World;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+// :::1
+@Mixin(ArrowEntity.class)
+public class ArrowEntityMixin {
+
```
Octol1ttle:
```suggestion
```

Octol1ttle (changes requested):
In addition to these comments, capitalization needs to be corrected (`fabric` -> `Fabric`, `mixin` -> `Mixin`)

IMB11:
I think the best course of action would be to move this page into an "Advanced Techniques" section that can in the future contain information on access wideners, accessors, duck interfaces, etc.

PR #119, opened 2024-06-20
404 page, localisation improvements

its-miroma:
- Implement 404 page
- Add more localisation strings
- Show page authors who contributed outside GitHub
- Update to latest `vitepress-versioning-plugin` version
- Update scrollbar colors
- Move `LICENSE.md` to `LICENSE`
- Improve `contributing.md` file
- Fix miscellaneous Markdown issues
  - Fix weird escapes in some translated links
  - `Codec&amp;amp;lt;Map&amp;amp;gt;` is getting out of hand
- Add `markdownlint-rule-search-replace` rules
  - `missing-heading-anchor`: All headings must have anchors
  - `no-absolute-links`: No absolute links allowed
  - `no-curly-quotes`: No curly quotes allowed
  - `no-file-extension-in-links`: Links should not have an extension
  - `no-multiple-spaces`: Yeah, no multiple spaces,
  - `no-punctuation-in-links`: Punctuation should not be clickable
  - `no-trailing-whitespace`: Extend `MD009` to code blocks

PR #118, opened 2024-06-19
Blocks

IMB11:
- [x] Creating Your First Block
- [x] Block States

Block entities and more complex stuff can be done after updating to 1.21, they change too much for it to be worth making a guide.

PR #116, opened 2024-06-14
Item pages

IMB11:
- [x] "Creating Your First Item"
	- [x] Change Screenshot
	- [x] Crafting Recipes
	- [x] Stack Size
	- [x] Compostable/Fuel
- [x] "Food Items"
- [x] "Armor"
	- [x] Knockback Resistance stuff
- [x] "Tools"
- [x] "Custom Items"
	- [x] Tooltips
- [x] "Custom Item Groups"
- [x] Text And Translations (along for the ride)

Also implemented:

- asset downloading components, self explanatory, allows users to download example textures etc.
- medium-zoom, allows users to zoom in images within the page, useful for pages with IDE screenshots and just in general for accessiblity.
- VidStack based videoplayer, will be useful for guides with webm videos, no more need to use audioless webp animations to show something in-game. It automatically has support for Youtube Embeds as well which will prove useful for #111 and future PRs with external videos.

PR #121, opened 2024-06-23
Change Identifier constructor to Identifier.of()

GraceHopper0012:
Minecraft has privatised the constructors of net.minecraft.util.Identifier in 1.21, have to use Identifier.of() instead


IMB11:
The docs pages are for 1.20.4, we will be updating them all at once after #116, #118 and #119 is merged.

IMB11:
docs pages*

PR #125, opened 2024-07-02
1.21

IMB11:


IMB11:
Should be good for review now, please have a thorough read of the pages that have changed:

- Potions
- Armor Materials
- Basic Rendering Concepts
- HUD rendering
- Text and Translations

Other than that, it's been pretty easy to update everything to 1.21, the versioning system works like a charm and the script works great at automating the bulk of the process. This only took 2ish hours, which is good considering it's the first time we've updated to a new version.

modmuss50:
Question: How is translation going to work? I assume the translations for the older versions are now set in stone?

modmuss50:
I like the warning / "This page is written for:" card on the older versions, not sure it really adds much on the latest version? Maybe could be a bit smaller or something, its very much a nit pick. Its great to see this system working 👍 Ill take a proper look at this PR soon.

IMB11:
Yup, they're set in stone, we could enable translations for versioned pages but I don't see a way on how to configure it properly in the crowdin.yml.

Also, if translations of versioned content is enabled, whenever we update the wiki, the crowdin string count would pretty much double, and thats not really good, as it lags the crowdin editor and you'd have to constantly bug the crowdin team for an open source string limit increase.

IMB11:
> I like the warning / "This page is written for:" card on the older versions, not sure it really adds much on the latest version? Maybe could be a bit smaller or something, its very much a nit pick. Its great to see this system working 👍 Ill take a proper look at this PR soon.

It's mostly temporary until I can develop an actual version switcher component. The next vitepress update is adding support for custom components within navigation bars which should allow me to do this. I agree it could be a bit smaller on latest pages though.

modmuss50:
> Yup, they're set in stone

Ok, great thats what I was going to suggest, I was worried about the string count. (I am yet to send that email :D) If there are issues or mistakes that urgently need fixing that can be done via github.

modmuss50 (dismissed):
Impossible to review, I trust this is ok. It seems so at a glance. Lets merge and fix any issues in other PRs. Great work 👍 

Going forward we can try to keep up with updates.

modmuss50 (approved):

PR #128, opened 2024-07-05
Recursive codecs name change

Rubyboat1207:
``Codecs#createRecursive`` does not exist anymore, and has been replaced with ``Codec#recurisve``. Not sure when this change exactly occurred on the Minecraft Codebase, however I have updated the docs to reflect this.

All instances of ``Codecs#createRecursive`` have been replaced with ``Codec#recurisve``, along with instances in any code examples. The usage of these functions is the same, so no extra changes were required. Additionally, I made sure not to update any 1.20.4 articles, as this was probably correct at the time.

### DIFF
--- a/develop/codecs.md
+++ b/develop/codecs.md
@@ -411,7 +411,7 @@ Our new codec will serialize beans to json like this, grabbing only fields that 
 
 ### Recursive Codecs {#recursive-codecs}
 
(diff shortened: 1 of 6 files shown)


IMB11 (approved):

IMB11:
Seems like Mojang updated DFU separately from Minecraft (we've seen them do this with the likes of authlib and brigadier), which is why this wasn't originally caught as an issue. Thanks.

PR #130, opened 2024-07-07
Fix #129

### LINKED ISSUE #129: Path mistake in new docs for Minecraft 1.21
On the new docs for Minecraft 1.21 for creating a crafting recipe [here](https://docs.fabricmc.net/develop/items/first-item#adding-a-basic-crafting-recipe), it states that the path to add recipes is `data/<mod id here>/recipes` when, in reality, Mojang changed it to `data/<mod id here>/recipe`. It was likely part of their recent renaming of package names to remove plurals.

megabyte6:
Change to the correct path for modding in Minecraft version 1.21


IMB11 (approved):

IMB11:
Thank you :+1:

PR #111, opened 2024-06-13
"Using the ide" Page [Part 1]

JR1811:
**Original page:** https://fabric.moddedmc.wiki/getting-started/using-the-ide
**Discussion in the Discord:** https://discord.com/channels/507304429255393322/1208846408552030238/1235450767226437692

## Content

I modified the content slightly to adapt to the new UI, mentioned split sources and added a small recommendation for the decompilers. Besides that, everything else should be mostly the same.

There are no changes in the `reference` directory since this topic doesn't focus on the code but rather on the usage and toolsets of the IDE. There shouldn't be any code logic which needs to be explained further or maintained in the future. They are only examples.

## Problems

The ["Basic Problem solving"](https://fabric.moddedmc.wiki/getting-started/basic-problem-solving) page is linked already, but it's not in the new docs yet. I'll add it to a later pull request asap, if this pull request gets accepted.

Also, VSCode displays the `MD033/no-inline-html: Inline HTML [Element: iframe]` warning on the iframe HTML element at the end of the new page. Is there a better way to implement videos like that?

PR #145, opened 2024-07-18
"Using the ide" Page [Part 2]

JR1811:
# Information

This PR is a continuation of https://github.com/FabricMC/fabric-docs/pull/111.

The Fork branch needed to be cleaned due to an error on my side, so here is the fresh PR with basically the same commits.

In addition, I also committed the changes suggested by @modmuss50 from back in the old PR (moved page to misc group)


develop/ide-tips-and-tricks.md
```diff
@@ -0,0 +1,246 @@
...
+authors: 
+    - JR1811
+---
+
+# IDE Tips and Tricks{#ide-tips-and-tricks}
+
+This page gives useful bits of information, to speed up and ease the workflow of developers. Incorporate them into yours, to your liking.
+It may take some time to learn and get used to the shortcuts and other options. You can use this page as a reference for that.
+
+::: note
```
IMB11:
```
::: warning
```

IMB11 (changes requested):
I think this is ready to merge, just need you to fix this one callout block. Only `info`, `warning` , `error`, `tip` and `details` are supported.

JR1811:
By the way, I've noticed that the vidstack video embed's thumbnails are massively zoomed in. Is there a way to fix it?
If not or it's too bothersome, it is good enough for me.

grafik

IMB11:
> By the way, I've noticed that the vidstack video embed's thumbnails are massively zoomed in. Is there a way to fix it? If not or it's too bothersome, it is good enough for me.
> 
> grafik

I think it's actually the first frame of the video, there's not much we can do about it, but it's a huge improvement over plain youtube iframes in terms of privacy and performance.

IMB11 (approved):

PR #148, opened 2024-07-20
Refactor Commands Guide

### LINKED ISSUE #94: develop/commands/basics points to non-existing code
In the _Command Redirects_ section there is a code box that currently displays "No lines matched." because it doesn't point to any existing code. The code example should be something using `redirect` on one of the previously registered commands. There should also be a paragraph explaining that you don't necessarily need to redirect to the root node, like in the original wiki article.

dicedpixels:
This may be a bit controversial 😉.

- Refactored the commands so that the `executes` logic is in a separate method. A method reference is then passed.
- I felt the use of the tater examples to be quite confusing (Yes, I wrote it). Hopefully, the new command literals are much more meaningful.
- Reorganized the `FabricDocsReferenceCommands.java` class. It now uses named transcludes. The numbers were hard to keep track of.
- Updated the images to be consistently sized.
- Add an info box for the usage of method references.
- Add a warning box for Command Redirects to highlight Brigadier's inconsistent behavior (https://github.com/Mojang/brigadier/issues/46).
- Fix command suggestion example to actually have a successful execution.
- Tested all the commands/arguments/suggestion providers.

What was not done:
- Break commands into nodes and then chaining them. Why? Despite being quite popular, I believe it's better for us to stick closer to the Vanilla/Brigadier way. Using method references seems to be a good middle ground for simpler commands like the doc's. If the nodes approach should be documented, it can be a separate page.
- Port changes to 1.20.4. This can be done later, as the diff was just the path to the reference mod.


IMB11:
Don't bother backporting to 1.20.4, it will break translations which are fixed for older versions. Generally older versions should only receive minor fixes like typos and syntax issues instead of full rewrites :)

IMB11:
@dicedpixels is this considered ready for review?

dicedpixels:
Yep!

IMB11 (approved):
This is pretty great, a much needed improvement.

PR #152, opened 2024-07-23
Fix armor items for 1.21

### LINKED ISSUE #151: Custom Armor guide leads to stackable, non-damageable armor items
**Docs version:** 1.21

If you follow the guide to create the Guidite Armor, the resulting items have no durability, which makes them stackable to 64 and take no damage when the player is hit. The item initializer uses the default-constructed `Item.Settings()` which do not include a `maxDamage` modifier. 

I believe in general that the durability for armor should be explained better, because it is quite unintuitive in the regard that you need a custom armor material which, unlike its `ToolMaterial` counterpart does not have any durability information. Therefor the durability of armor items has to be custom-set and is usually (e.g. vanilla items) derived from a general "material durability" and a specific "slot durability" for each armor slot. 

I think it would be good to have a small section that explains why you need to manually set the durability and some best-practices on calculating the value (material durability * slot durability).

DWalz:
Minecraft changed the way durability works with armor items and armor materials. I added a small explainer to durability in the docs.

Closes #151 

Note regarding [the implementation](https://github.com/FabricMC/fabric-docs/issues/151#issuecomment-2246285823): I used a durability multiplier and referenced the `ArmorItem.Type`'s base durabilities mimicking the vanilla armor registries.

### DIFF
--- a/develop/items/custom-armor.md
+++ b/develop/items/custom-armor.md
@@ -95,11 +95,11 @@ Obviously, an armor set doesn't need every type to be satisfied, you can have a 
 
 ### Durability {#durability}
 
-Unlike `ToolMaterial`, `ArmorMaterial` doesn't store any information about the durability of items.
-For this reason the durability needs to be added to the armor items' `Item.Settings` when registering them.
+Unlike `ToolMaterial`, `ArmorMaterial` does not store any information about the durability of items.
+For this reason the durability needs to be manually added to the armor items' `Item.Settings` when registering them.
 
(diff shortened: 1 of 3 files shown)


DWalz:
Let me know if the wording is clear enough and the solution I used good enough.

IMB11 (dismissed):

IMB11:
Just a single lint issue.

DWalz:
I should have fixed the linting error. I don't know why it came up locally because I did test for it but I probably made some mistake. Thanks for your time!

IMB11 (approved):

PR #153, opened 2024-07-25
Fix contribution guidelines link to 404

kevinthegreat1:
Self explanatory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,4 +4,4 @@ This repository contains the source code for the Fabric Documentation website, w
 
 ## Contributing
 
-Please refer to the [contribution guidelines](./contributing) for more information.
+Please refer to the [contribution guidelines](./contributing.md) for more information.


its-miroma (commented):
This link shouldn't be broken, where did it not work?

https://docs.fabricmc.net/contributing#use-relative-links

kevinthegreat1:
> This link shouldn't be broken, where did it not work?
> 
> https://docs.fabricmc.net/contributing#use-relative-links

The github homepage. <https://github.com/FabricMC/fabric-docs>  
It's also broken in <https://github.com/FabricMC/fabric-docs/blob/main/README.md>

its-miroma:
None of the links work in GitHub...

cc. @IMB11

IMB11:
need linting to be excluded for contributing file then

its-miroma:
Readme is already excluded. All clear!

its-miroma (approved):

PR #157, opened 2024-08-06
Fix editor config import layout to match spotless

kevinthegreat1:
Fix editor config currently not matching spotless.

I've tested this in every java file in the reference mod and it works without any problems.


modmuss50 (approved):
Thanks

PR #160, opened 2024-08-07
Add guides on using Fabric with Third Party Launchers

IMB11:
**Need someone to write the modrinth app guide, I was unable to launch the production version it on linux due to having a dev version of it installed.**

**Need someone to write the CurseForge Launcher guide, the linux version is not the same as the windows/mac versions.**

These pages simply tell users:

- How to install a Fabric Loader instance within the launcher.
- How to install Fabric mods with the launcher.

Nothing more.

The following launchers are included, mostly because they're also on the redesign PR for the website (https://github.com/FabricMC/fabricmc.net/pull/103)

- [x] GDLauncher (Carbon)
- [x] Prism Launcher
- [x] FTB App
- [ ] CurseForge App
- [ ] Modrinth App
- [x] ATLauncher


modmuss50 (changes requested):
I think this would be better suited to the wiki. This is becuase we dont endorse or provide support for these launchers, e.g we cannot guarenteee the security of them. Its not fair on the other launchers that arent on here, on the wiki they can add a page them selves. I would expect most users starting out to use the vanilla launcher.

IMB11:
> This is because we dont endorse or provide support for these launchers

Yeah but we also have to acknowledge that there is a wide range of users using these launchers - if this was the case, there shouldn't be any third-party launcher content on the wiki either.

> on the wiki they can add a page them selves.

They can just make a PR here?

modmuss50:
> we also have to acknowledge that there is a wide range of users using these launchers
Yes, this can be done by linking to the wiki

The wiki is better suited for 3rd part launchers like this, the docs site is supposed to cover the official way of installing fabric via the vanilla launcher, and it does that well. Its not our job to try and maintain high quality docs on other projects.

IMB11:
I've added

```
::: warning
Fabric does not endorse any third-party tools listed within this section, and we do not provide support for them. If you encounter any issues with these tools, please report them to the respective tool's developers.
:::
``` 

to the pages, which should be suitable enough to explain that we don't endorse third-party launchers.

IMB11:
> > we also have to acknowledge that there is a wide range of users using these launchers
> > Yes, this can be done by linking to the wiki
> 
> The wiki is better suited for 3rd part launchers like this, the docs site is supposed to cover the official way of installing fabric via the vanilla launcher, and it does that well. Its not our job to try and maintain high quality docs on other projects.

I agree, but it would provide some well needed consistency to these pages - the wiki pages are outdated and need full replacement, they also require much more maintenance due to the fact they're telling users how to install said launchers.

PR #167, opened 2024-08-17
Custom Resources Article

Manchick0:
Introduced a comprehensive guide to custom resources, covering client and server-side data.


develop/custom-resources.md
```diff
@@ -0,0 +1,126 @@
...
+
+> We'll create a `JsonDataLoader` for loading JSON files from the `fruit` folder
+> in a data pack, deserialize them into actual fruits, and store them
+> for further retrieval.
+
+Let's start by creating a class itself. It should extend `JsonDataLoader`, and override the constructor
+and the `apply` method.
+
+```java
+public class FruitDataLoader extends JsonDataLoader {
```
IMB11: Needs to be in the reference mod. Just because it uses mc classes which may change in the future (ResourceManager)

IMB11 (changes requested):
Really good, just a minor nitpick.

develop/custom-resources.md
Manchick0: Sure! I added a new class, which is just empty. I really didn't want to do it though, that's why I decided to use raw code in the previous version.

Manchick0:
Any updates?

PR #174, opened 2024-08-27
SoundInstances page

### LINKED ISSUE #70: Developer Guides: SoundInstances
This would fit well into the `Sounds` section being added by #68

JR1811:
From Issue #70

# Content

This page is split into several parts.

1. again, preparing sound files for SoundInstances
2. using basic SoundInstances
3. Advanced SoundInstance usage example

The page first explains how to create and use simple SoundInstances and then shows an example of how to make a sound system, which can modulate e.g. its volume and pitch based on the current sound phase and connected BlockEntity values.

# Problems & Questions

The current page has not been cleaned up and spellchecked yet. The main reason I'm PRing this now is to ask for opinions.
I was about to delete the 3rd section _(starting from the  `Advanced SoundInstances` header)_ because of the scope of this topic. But it shows how abstract systems can be created. I'm pretty sure this "huge" in-depth example does not belong in the Fabric Docs, but I wanted to hear other opinions before I scrap all of that again.

**The Choices before I move into the clean-up phase:**

- keep the current example in (with clean-up etc. afterwards)
- massively shrink down / rework the example by losing most of the abstractions and inheritance
- take the whole example out, only the basic SoundInstance usage is needed

---

This page uses but does not cover BlockEntities and custom S2C networking. Those should be handled on separate pages.


Tewxx:
I this is is well written and should be added

JR1811:
Just wondering. Has anyone checked this yet?
If it's in a usable state, ill clean it up / continue working on it.

Otherwise, we can just get rid of this page.

IMB11:
This is incredibly well made.

IMB11 (dismissed):

IMB11:
Please can you fix the spotless issues and markdownlint issues present, and then LGTM.

its-miroma (approved):

its-miroma:
LGTM @IMB11

IMB11 (approved):

PR #177, opened 2024-08-28
Path needs to be recipes instead of recipe

Bedlessdeus:
Updated path to be the correct one, or else, the game wont load said recipe


Juuxel:
The directory name is singular in MC 1.21, so the docs are correct.

Bedlessdeus:
Could a note be added for the versions under 1.20 that require this to be plural?

Linguardium:
> Could a note be added for the versions under 1.20 that require this to be plural?

click the version dropdown and choose 1.20.4

Bedlessdeus:
Didn't see that before, sorry!

PR #182, opened 2024-09-17
Additional translatable text information

LordEnder-Kitty:
Added additional information about translatable text


IMB11 (approved):

its-miroma (approved):
Approving to run checks

PR #183, opened 2024-09-19
Update automatic-testing.md

dozed-dev:
Gametest framework is exclusive to Bedrock Edition, so it probably shouldn't be mentioned in Fabric's automated testing guide

### DIFF
--- a/develop/automatic-testing.md
+++ b/develop/automatic-testing.md
@@ -1,21 +1,13 @@
 ---
 title: Automated Testing
-description: A guide to writing automatic tests with Fabric Loader JUnit.
+description: A guide to writing automatic unit tests with Fabric Loader JUnit.
 authors:
   - kevinthegreat1
 ---
 
 # Automated Testing {#automated-testing}
 
-This page explains how to write code to automatically test parts of your mod. There are two ways to automatically test your mod: unit tests with Fabric Loader JUnit or game tests with the Gametest framework from Minecraft.
-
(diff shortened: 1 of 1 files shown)


modmuss50:
> Gametest framework is exclusive to Bedrock Edition

Thats not true, Java edition has a fully featured test framework.

dozed-dev:
> > Gametest framework is exclusive to Bedrock Edition
> 
> Thats not true, Java edition has a fully featured test framework.

Sorry :sweat_smile:  
I'm currently learning about tests for Minecraft mods, and most results in the search engine about "Minecraft GameTest framework" is about the gametest for Bedrock edition (like [here](https://www.minecraft.net/en-us/creator/article/get-started-gametest-framework) and [here](https://learn.microsoft.com/en-us/minecraft/creator/documents/gametestgettingstarted)). I couldn't find any info on Java edition gametest, and on [Minecraft wiki](https://minecraft.wiki/w/Commands/gametest), it says that /gametest command is exclusive to bedrock edition. So I got confused and assumed that the docs are wrong and it's exclusive to bedrock.

The only info I eventually found about Fabric's gametest API is in its [source code docs](https://github.com/FabricMC/fabric/blob/1.21.1/fabric-gametest-api-v1/src/main/java/net/fabricmc/fabric/api/gametest/v1/package-info.java#L18) and in Minecraft source code mappings (net.minecraft.test).

Sorry again for the wrong pull request!

PR #194, opened 2024-10-15
Update labels in `contributing.md`

### LINKED ISSUE #193: Update contributing.md to reflect label rename

its-miroma:
Fixes #193

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -19,15 +19,18 @@ All contributions must follow our [style guidelines](#style-guidelines).
 
 If you want to translate the documentation into your language, you can do this on the [Fabric Crowdin page](https://crowdin.com/project/fabricmc).
 
-## Contributing Content {#contributing-content}
+<!-- markdownlint-disable-next-line titlecase-rule -->
+## <Badge type="tip">new-content</Badge> Contributing Content {#contributing-content}
 
 Content contributions are the main way to contribute to the Fabric Documentation.
 
-All content contributions go through three stages:
+All content contributions go through the following stages, each of which is associated with a label:
 
-1. Guidance for Expansion (if needed)
-2. Content Verification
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
   },
   "devDependencies": {
     "@types/glob": "^8.1.0",
-    "@types/node": "^22.5.5",
+    "@types/node": "^22.7.5",
```
modmuss50: Was there a reason to update the depedencies?
Shouldnt the package-lock file also changed?

its-miroma: I had to add `@vueuse/core`, because `vitepress dev` failed without it.

its-miroma: I downgraded all other dependencies

modmuss50: You should be using `npm run dev`, Im always very hesitent to add new depedencies like this. I will take a closer look.

modmuss50: That works for me, I dont see why this package should be added.

its-miroma: `@vueuse/core` is being used here: https://github.com/FabricMC/fabric-docs/blob/main/.vitepress/theme/components/BannerComponent.vue#L2

its-miroma: `npm run dev` does not fail, but opening the page in the browser does. I'll attach a screenshot

modmuss50: As you can see from the package lock changes `@vueuse/core` was already in there as vitepress depends on it. I dont think this PR makes any functional changes to how it works.

its-miroma: Honestly I can't reproduce the problem I had yesterday, so I guess it's fixed (:

modmuss50: Humm, do let me know if that changes. Sorry about being so cautious when adding dependepes, especially npm ones :D

modmuss50 (approved):

PR #202, opened 2024-10-24
Fix indentation for sounds.json

cprodhomme:


its-miroma (approved):

its-miroma:
Yeah, we have actions to format Java and Markdown; we may want to format JSON and other languages...

PR #242, opened 2024-12-30
Update to 1.21.4

modmuss50:
Supersedes #206, #210, #226

Fixes #216, #231, #233

Depends on #244

---

### TODO

- [x] The mod itself (@modmuss50) #243
- [x] Introduction to Fabric and Modding (@IMB11) #245 
- [x] Setting up a Development Environment (@IMB11) #245 
- [x] Creating a Project (@IMB11) #245 
- [x] Project Structure (@IMB11) #245 
- [x] Launching the Game (@IMB11) #245 
- [x] Creating Your First Item (@IMB11) #253
- [x] Food Items (@IMB11) #253
- [x] Custom Tools (@IMB11) #253
- [x] Custom Armor (@IMB11) #253 
- [x] Custom Item Groups (@IMB11) #253
- [x] Custom Item Interactions (@IMB11) #253
- [x] Custom Enchantment Effects (@IMB11) #253
- [x] Custom Data Components (@IMB11) #253
- [x] Potions (@IMB11) #253
- [x] Creating your first block (@IMB11) #245 
- [x] Block States (@IMB11) #245 
- [x] Block Entities (@IMB11) #245 
- [x] Block Entity Renderers (@IMB11) #245 
- [x] Status Effects
- [x] Damage Types (@IMB11) #258
- [x] Playing Sounds
- [x] Creating Custom Sounds
- [x] Dynamic Sounds (Sound Instances)
- [x] Creating Commands
- [x] Command Arguments
- [x] Command Suggestions
- [x] Basic Rendering Concepts (@IMB11) #248
- [x] Using the Drawing Context (@IMB11) #248
- [x] Rendering in the HUD (@IMB11) #248
- [x] Custom Screens (@IMB11) #248
- [x] Custom Widgets (@IMB11) #248
- [x] Creating Custom Particles (@IMB11) #248
- [x] Data Generation Setup
- [x] Tag Generation
- [x] Translation Generation
- [x] Advancement Generation (@IMB11) #258 
- [x] Recipe Generation
- [x] Loot Table Generation
- [x] Codecs
- [x] Events
- [x] Text and Translations
- [x] IDE Tips and Tricks
- [x] Automated Testing

### Existing Large Changes

These changes will need to be taken into consideration when updating their pages.

- [x] Advancement criteria in data generator now use registry wrappers.
- [x] Renderlayers must be passed to drawTexture to draw textures on screens.
- [x] Shaders moved from GameRenderer to ShaderProgramKeys
[...]

PR #248, opened 2024-12-30
Update to 1.21.4: "Rendering" 

### LINKED ISSUE #173: "Tick delta" explanation and usage are incorrect
Tick delta is a term used quite often in the rendering pages, The [Basic Rendering Concepts](https://docs.fabricmc.net/develop/rendering/basic-concepts) page refers to it as "which is the time that has passed since the last frame". [Rendering in the Hud](https://docs.fabricmc.net/develop/rendering/hud) further elaborates on this by explaining it as "the time since the last frame, in seconds."

As obvious from the issue title, both of these explanations are wrong. Unfortunately, not everyone is aware of this. Browsing the `mod-dev` channels in the Fabricord, it is quite easy to find examples of people sharing this wrong explanation, and other people encountering issues in their code when they use it as explained in the docs or by others.

Let's look at the game code to determine what "tick delta" actually is:
```java
public int beginRenderTick(long timeMillis) {
		this.lastFrameDuration = (float)(timeMillis - this.prevTimeMillis) / this.targetMillisPerTick.apply(this.tickTime);
		this.prevTimeMillis = timeMillis;
		this.tickDelta = this.tickDelta + this.lastFrameDuration;
		int i = (int)this.tickDelta;
		this.tickDelta -= (float)i;
		return i;
	}
```
This is Yarn mappings. Unfortunately, the `lastFrameDuration` field name is also incorrect, I'll PR a fix some time later.
Looking at this, we see that `tickDelta` gets incremented with `lastFrameDuration` (which is the time between frames, converted to game ticks, depending on the server's selected tick rate), and then `i` is subtracted from it. `i` is calculated by casting the `tickDelta` to an integer, which drops the decimal part. So converting `319.64` would result in `319`, `59.03` would result in `59`.

Two things are obvious here:
1) the calculation of `tickDelta` on the current frame is dependent on the value from the previous frame
2) `tickDelta` is never less than 0 or more than 1

With this mind, we can come to the conclusion that `tickDelta` is the "progress" between the last game tick and the next game tick. For example, if we assume the game runs at 60 FPS, this is what `tickDelta` would be on different frames:
* Frame 1: game tick, tickDelta is 1.0
* Frame 2: no game tick, tickDelta is 0.33
* Frame 3: no game tick, tickDelta is 0.67
* Frame 4: game tick, tickDelta is 1.0

We can prove this further by looking at usages of `tickDelta`. `tickDelta` is often used in combination with `MathHelper#lerp`, which would never work correctly if `tickDelta` was actually time since last frame.

I hope this is sufficient evidence that the explanation needs to be corrected. The implementation of `RenderTickCounter#getTickDelta` can be looked at if this is not enough proof

IMB11:
This PR verifies and updates the pages in the "Rendering" category.

Closes #173


develop/rendering/hud.md
```diff
@@ -7,23 +7,38 @@ authors:
...
 
-To register to this event, you can simply call `HudRenderCallback.EVENT.register` and pass in a lambda that takes a `DrawContext` and a `float` (deltaTick) as parameters.
+To register to this event, you can simply call `HudRenderCallback.EVENT.register` and pass in a lambda that takes a `DrawContext` and a `float` (tickDelta) as parameters.
 
 The draw context can be used to access the various rendering utilities provided by the game, and access the raw matrix stack.
 
 You should check out the [Draw Context](./draw-context) page to learn more about the draw context.
 
-### DeltaTick {#deltatick}
+### Tick Delta {#tickdelta}
```
its-miroma:
```suggestion
### Tick Delta {#tick-delta}
```

its-miroma:
cc. @Octol1ttle

develop/rendering/hud.md
```diff
@@ -7,23 +7,38 @@ authors:
...
 
-We already briefly touched on rendering things to the hud in the [Basic Rendering Concepts](./basic-concepts) page and [Using The Drawing Context](./draw-context), so on this page we'll stick to the `HudRenderCallback` event and the `deltaTick` parameter.
+We already briefly touched on rendering things to the hud in the [Basic Rendering Concepts](./basic-concepts) page and [Using The Drawing Context](./draw-context), so on this page we'll stick to the `HudRenderCallback` event and the `tickDelta` parameter.
 
 ## HudRenderCallback {#hudrendercallback}
 
 The `HudRenderCallback` event - provided by Fabric API - is called every frame, and is used to render things to the HUD.
 
-To register to this event, you can simply call `HudRenderCallback.EVENT.register` and pass in a lambda that takes a `DrawContext` and a `float` (deltaTick) as parameters.
+To register to this event, you can simply call `HudRenderCallback.EVENT.register` and pass in a lambda that takes a `DrawContext` and a `float` (tickDelta) as parameters.
```
Octol1ttle: In 1.21.4, `HudRenderCallback` provides a `RenderTickCounter` instead of a `float`

develop/rendering/hud.md
```diff
@@ -7,23 +7,38 @@ authors:
...
+| 1     | `1.0` (new tick)                             |
+| 2     | `0.11 (1÷9)` - The next tick is in 9 frames. |
+| 3     | `0.22 (2÷9)`                                 |
+| 4     | `0.33 (3÷9)`                                 |
+| 5     | `0.44 (4÷9)`                                 |
+| 6     | `0.55 (5÷9)`                                 |
+| 7     | `0.66 (6÷9)`                                 |
+| 8     | `0.77 (7÷9)`                                 |
+| 9     | `0.88 (8÷9)`                                 |
+| 10    | `1.0 (9÷9)` (new tick)                       |
```
Octol1ttle: This section was confusing to me when I first read it. The first sentence assumes that the game is running at 60 FPS, but the table uses 200 FPS

develop/rendering/hud.md
```diff
@@ -7,23 +7,38 @@ authors:
...
+
+Practically, let's say you want to lerp a color over time. You can use the `tickDeltaManager` to get the tickDelta, and store it over time to lerp the color:
```
Octol1ttle: Is this a Mojmap name? It's called `RenderTickCounter` in Yarn

develop/rendering/hud.md
Octol1ttle, line 1: https://github.com/FabricMC/fabric-docs/blob/port/1.21.4/reference/latest/src/client/java/com/example/docs/rendering/HudRenderingEntrypoint.java

`totalTickDelta` is still a flawed way to lerp a color like this. It will *appear* to work fine at low FPS, but at higher FPS (120+) it will look very broken (for reasons I shouldn't have to explain here)

The best way for that example to work is to use `Util.getMeasuringTimeMs()` to make a `timeDelta` and have a `totalTimeDelta`.

If we want to show the importance of `tickDelta`, we need something that is tick-based. A good example would be rendering text above the entity (similar to how it's done in the Block Entity Renderers article). Without using `tickDelta`, the text will jitter.

Octol1ttle (changes requested):

develop/rendering/hud.md
IMB11: It's the name of the parameter passed by the event

develop/rendering/hud.md
IMB11: Ah, I forgot to update the FPS

Octol1ttle (approved):
Looks fine to me.

its-miroma (approved):
The linter will run once this is merged into #242

PR #249, opened 2024-12-31
Networking Guide

dicedpixels:
Ports [Networking](https://wiki.fabricmc.net/tutorial:networking).

Differences:

- Actual working item, Lightning Tater.
- Along with S2C, a C2S example is also included.


reference/latest/src/main/resources/fabric.mod.json
```diff
@@ -19,16 +19,19 @@
...
 		"com.example.docs.client.command.FabricDocsReferenceClientCommands",
-		"com.example.docs.FabricDocsDynamicSound"
+		"com.example.docs.FabricDocsDynamicSound",
+		  "com.example.docs.network.basic.FabricDocsReferenceNetworkingBasicClient"
 	  ],
 	  "fabric-datagen": [
 		"com.example.docs.FabricDocsReferenceDataGenerator",
-		"com.example.docs.damage.FabricDocsReferenceDamageTypesDataGenerator"
+		"com.example.docs.damage.FabricDocsReferenceDamageTypesDataGenerator",
+		"com.example.docs.networking.basic.FabricDocsReferenceNetworkingBasicDataGenerator"
```
IMB11: You cant have multiple data generators, it doesnt work. Please use the main datagenerator (FabricDocsReferenceDataGenerator)

reference/latest/src/main/java/com/example/docs/networking/basic/FabricDocsReferenceNetworkingBasicDataGenerator.java
```diff
@@ -0,0 +1,33 @@
...
+import net.minecraft.data.client.BlockStateModelGenerator;
+import net.minecraft.data.client.ItemModelGenerator;
+import net.minecraft.data.client.Models;
+
+import net.fabricmc.fabric.api.datagen.v1.DataGeneratorEntrypoint;
+import net.fabricmc.fabric.api.datagen.v1.FabricDataGenerator;
+import net.fabricmc.fabric.api.datagen.v1.FabricDataOutput;
+import net.fabricmc.fabric.api.datagen.v1.provider.FabricModelProvider;
+
+public class FabricDocsReferenceNetworkingBasicDataGenerator implements DataGeneratorEntrypoint {
```
IMB11: Can't have multiple data generators. See https://github.com/FabricMC/fabric/issues/4330

IMB11 (changes requested):

reference/latest/src/main/resources/fabric.mod.json
IMB11: Move it into the main sourceset if required.

IMB11: If you do add a provider to the main data generator, add it to the end of the list, some pages are referencing the file by line number to show the `pack.addProvider` call. 

Example:

chrome_bVhPXuS4Yq

develop/networking.md
```diff
@@ -0,0 +1,255 @@
...
+
+`S2C` and `C2S` are two common suffixes that mean _Server-to-Client_ and _Client-to-Server_ respectively.
+
+### Sending a Packet to the Client {#sending-a-packet-to-the-client}
+
+To send a packet with our custom payload, we can use `ServerPlayNetworking.send` which takes in a `ServerPlayerEntity`
+and a `CustomPayload`.
+
+Let's start by creating our Lightning Tater item. You can override `use` to add trigger an action when the item is used.
+In this case, let's send packets to all the players in the server.
```
Octol1ttle: This statement is wrong. In the example code, the payload is only sent to players tracking the `BlockPos`. Tracking is a very important concept for efficient networking, it shouldn't be omitted.

develop/networking.md
```diff
@@ -0,0 +1,255 @@
...
+
+We only send packets when the action is initiated on the server, by returning early with a `isClient` check:
+
+@[code lang=java transclude={23-25}](@/reference/latest/src/main/java/com/example/docs/networking/basic/LightningTaterItem.java)
+
+We create an instance of the payload with the user's position:
+
+@[code lang=java transclude={27-27}](@/reference/latest/src/main/java/com/example/docs/networking/basic/LightningTaterItem.java)
+
+Finally, we get all the players in the server through `PlayerLookup` and send a packet to each player.
```
Octol1ttle: Ditto

develop/networking.md
```diff
@@ -0,0 +1,255 @@
...
+Then we register our payload in our **common initializer**. However, this time as _Client-to-Server_ payload by using
+`PayloadTypeRegistry.playC2S().register`.
+
+@[code lang=java transclude={22-22}](@/reference/latest/src/main/java/com/example/docs/networking/basic/FabricDocsReferenceNetworkingBasic.java)
+
+To send a packet, let's add an action when the player uses a Poisonous Potato. We'll be using the `UseItemCallback`
+event to
+keep things concise.
+
+We register the event in our **client initializer**. This ensures that the action is only triggered on the client.
```
Octol1ttle: This isn't how sides work. A callback registered on the physical client can still be called on the logical server (when the client is an integrated server host)

develop/networking.md
```diff
@@ -0,0 +1,255 @@
...
+<VideoPlayer src="/assets/develop/networking/summon-lightning.webm" title="Summon lightning using Lightning Tater" />
+
+### Sending a Packet to the Server {#sending-a-packet-to-the-server}
+
+Just like sending a packet to the client, we start by creating a custom payload. This time, we will be sending the
+player's name and position to the server.
+
+@[code lang=java transcludeWith=::use_poisonous_potato_payload](@/reference/latest/src/main/java/com/example/docs/networking/basic/UsePoisonousPotatoPayload.java)
+
+Note the usage of `PacketCodec.tuple` to combine two values, a `String` and a `BlockPos`.
```
Octol1ttle: I know this is done for the sake of example, but this example is inefficient and dangerous:
1) Sending any data about the player is pointless because the server already knows which player sent it and can access any information it needs there.
2) If the client or server want to refer to another entity in a packet, they should use it's numerical ID (not UUID)
3) A malicious client can exploit this because the server doesn't check that any data is factually correct (data validation in C2S packets is also an important topic).

Octol1ttle (changes requested):

develop/networking.md
dicedpixels: 1. Yes, this I figured. I wanted a simple example.
2. I believe we can overlook this for the sake of simplicity.
3. This however is concerning. I was not aware and didn't think that far as we're dealing with the basics. I'm open for a better example that could address above 2 points as well.

develop/networking.md
dicedpixels: Noted. I believe using `world` would be better to match the description. Then I could add a statement about tracking and why it is important.

develop/networking.md
dicedpixels: ~~I'll remove this statement and move the callback registration to the logical side.~~

Edit: I don't think I understood this correctly. Any suggestions to rewrite this better?

develop/networking.md
dicedpixels: I've now changed the example to match what's written and added an info block about `PlayerLookup`.

develop/networking.md
dicedpixels: Addressed through previous comment.

develop/networking.md
dicedpixels: I believe adding a `isClient` check would restrict the behavior to the logical client. I've made this change. Hopefully, this is a better way of writing this example. If not, guide me through it.

develop/networking.md
Octol1ttle: If I were to write this, I would add a situation where a player is casting some sort of spell to another entity. In this situation, efficient networking can be showcased (see 2) and proper data validation can be showcased with checks like "does the entity actually exist and is not invulnerable?" "is the player close enough to the entity?"

reference/latest/src/client/java/com/example/docs/network/basic/FabricDocsReferenceNetworkingBasicClient.java
```diff
@@ -0,0 +1,56 @@
...
+					entity.setPosition(entityPos.getX(), entityPos.getY(), entityPos.getZ());
+					world.addEntity(entity);
+				}
+			}
+		});
+		// :::client_global_receiver
+
+		// :::use_item_callback
+		UseItemCallback.EVENT.register((playerEntity, world, hand) -> {
+			if (world.isClient) {
```
Octol1ttle:
```suggestion
			if (world.isClient()) {
```

reference/latest/src/main/java/com/example/docs/networking/basic/LightningTaterItem.java
```diff
@@ -0,0 +1,36 @@
...
+
+// :::lightning_tater_item
+public class LightningTaterItem extends Item {
+	public LightningTaterItem(Settings settings) {
+		super(settings);
+	}
+
+	@Override
+	public TypedActionResult<ItemStack> use(World world, PlayerEntity user, Hand hand) {
+		if (world.isClient) {
```
Octol1ttle:
```suggestion
		if (world.isClient()) {
```

develop/networking.md
```diff
@@ -0,0 +1,267 @@
...
+Then we register our payload in our **common initializer**. However, this time as _Client-to-Server_ payload by using
+`PayloadTypeRegistry.playC2S().register`.
+
+@[code lang=java transclude={22-22}](@/reference/latest/src/main/java/com/example/docs/networking/basic/FabricDocsReferenceNetworkingBasic.java)
+
+To send a packet, let's add an action when the player uses a Poisonous Potato. We'll be using the `UseItemCallback`
+event to
+keep things concise.
+
+We register the event in our **client initializer**. This ensures that the action is only triggered on the client.
```
Octol1ttle:
```suggestion
We register the event in our **client initializer** and we use `isClient()` to ensure that the action is only triggered on the logical client.
```

Octol1ttle (changes requested):

IMB11:
Hi! The documentation has been updated to version 1.21.4. Please rebase your PR and update your code to align with the latest version. 

Alternatively, you can create a new branch based on the latest commit from the [`FabricMC:main`](https://github.com/FabricMC/fabric-docs/tree/main) branch, copy over your changes from the old [`dicedpixels:networking`](https://github.com/dicedpixels/fabric-docs/tree/networking) branch, update it, and open a new PR - this approach is often simpler than dealing with complex git commands if you're not the best git wrangler.

Once you've completed the updates and are confident everything is ready, feel free to unmark this as a draft. Thanks again for your contribution!

develop/networking.md
dicedpixels: I've replaced the example with a simple instance of applying an effect on to a living entity, including some checks. Also added note on the importance of server side validation. Hopefully, it's enough to cover the points raised here while keeping it simple for the docs.

its-miroma:
Hello @dicedpixels, are you still working on this?

dicedpixels:
> Hello @dicedpixels, are you still working on this?

I guess so 😅. Haven't had time to sit and look back at it. I believe I've addressed all the comments, PR should be up-to-date. I will undraft it so more people can review.

develop/data-generation/loot-tables.md
```diff
@@ -18,7 +18,7 @@ Make sure you've completed the [datagen setup](./setup) process first.
 
 You will need different providers (classes) for blocks, chests, and entities. Remember to add them all to your pack in your `DataGeneratorEntrypoint` within the `onInitializeDataGenerator` method.
 
-@[code lang=java transclude={32-33}](@/reference/latest/src/client/java/com/example/docs/datagen/FabricDocsReferenceDataGenerator.java)
+@[code lang=java transclude={35-36}](@/reference/latest/src/client/java/com/example/docs/datagen/FabricDocsReferenceDataGenerator.java)
```
its-miroma:
```suggestion
@[code lang=java transclude={34-35}](@/reference/latest/src/client/java/com/example/docs/datagen/FabricDocsReferenceDataGenerator.java)
```

develop/networking.md
```diff
@@ -0,0 +1,277 @@
...
+    public HighlightingWandItem(Item.Settings settings) {
+        super(settings)
+    }
+
+    public TypedActionResult<ItemStack> use(World world, PlayerEntity user, Hand hand) {
+        // Raycast and find the block the user is facing at
+        BlockPos target = ...
+
+        // BAD CODE: DON'T EVER DO THIS!
+        ClientBlockHighlighting.highlightBlock(MinecraftClient.getInstance(), target);
```
its-miroma:
```suggestion
        // BAD CODE: DON'T EVER DO THIS! // [!code error]
        ClientBlockHighlighting.highlightBlock(MinecraftClient.getInstance(), target); // [!code error]
```

develop/networking.md
```diff
@@ -0,0 +1,277 @@
...
+Below code is for demonstration purposes only.
+:::
+
+Say you had a Wand which highlights the block you're looking, which will be visible to all
+nearby players:
+
+```java
+public class HighlightingWandItem extends Item {
+    public HighlightingWandItem(Item.Settings settings) {
+        super(settings)
```
its-miroma:
```suggestion
        super(settings);
```

develop/networking.md
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    public TypedActionResult<ItemStack> use(World world, PlayerEntity user, Hand hand) {
+        // Raycast and find the block the user is facing at
+        BlockPos target = ...
```
its-miroma:
```suggestion
        BlockPos target = ...
```

let's not give instructions on how to use this code (:

develop/networking.md
```diff
@@ -0,0 +1,277 @@
...
+    }
+}
+```
+
+Upon testing, you will see a lightning bolt being summoned and nothing crashes. Now you want to show the mod to your
+friend, you boot up a dedicated server and invite your friend on with the mod installed.
+
+You use the item and the server crashes. You will probably notice in the crash log an error similar to this:
+
+```text
```
its-miroma:
```suggestion
```log
```

develop/networking.md
```diff
@@ -0,0 +1,277 @@
...
+To fix this issue, you need to understand how Minecraft communicates between the game client and dedicated
+server.
+
+![Sides](/assets/develop/networking/sides.png)
+
+The diagram above shows that the game client and dedicated server are separate systems, bridged together using
+_packets_. Packets can contain data which we refer to as the _payload_.
+
+This packet bridge does not only exist between a game client and dedicated server, but also between your client and
+another client connected over LAN. The packet bridge is also present even in single-player. This is because the game
```
its-miroma:
```suggestion
another client connected over LAN. The packet bridge is also present even in singleplayer. This is because the game
```

develop/networking.md
```diff
@@ -0,0 +1,277 @@
...
+client will spin up a special integrated server instance to run the game on.
+
+The key difference between the three types of connections that are shown in the table below:
+
+| Connection Type               | Access to Game Client      |
+|-------------------------------|----------------------------|
+| Connected to dedicated Server | None (Server crash)        |
+| Connected over LAN            | Yes (Not host game client) |
+| Single-player (or LAN host)   | Yes (Full access)          |
```
its-miroma:
```suggestion
| Singleplayer (or LAN host)    | Yes (Full access)          |
```

develop/networking.md
```diff
@@ -0,0 +1,277 @@
...
+
+It may seem complicated to have communication with the server in three different ways. However, you don't need to
+communicate in three different ways with the game client. Since all three connection types communicate with the game
+client using packets, you only need to communicate with the game client like you're always running on a dedicated
+server.
+
+Connection to a server over LAN or single-player can also be treated like the server is a remote, dedicated server; so
```
its-miroma:
```suggestion
Connection to a server over LAN or singleplayer can also be treated like the server is a remote, dedicated server; so
```

develop/networking.md
```diff
@@ -0,0 +1,277 @@
...
+Before we send a packet with our custom payload, we need to register it.
+
+This can be done in our **common initializer** by `PayloadTypeRegistry.playS2C().register` which takes in a
+`CustomPayload.Id` and a `PacketCodec`.
+
+@[code lang=java transclude={28-28}](@/reference/latest/src/main/java/com/example/docs/networking/basic/FabricDocsReferenceNetworkingBasic.java)
+
+A similar method exists to register client-to-server payloads: `PayloadTypeRegistry.playC2S().register`.
+
+`S2C` and `C2S` are two common suffixes that mean _Server-to-Client_ and _Client-to-Server_ respectively.
```
its-miroma: should this appear earlier?

public/assets/develop/networking/sides.png
its-miroma, line 1: I would really prefer if this wasn't an image as it is difficult to modify in the future...

its-miroma (changes requested):

public/assets/develop/networking/sides.png
dicedpixels: Me too, but do we have any other option?

develop/networking.md
dicedpixels: I moved this to an info box just after "Defining a Payload" since from that point onwards there will be several usages of S2C and C2S methods.

develop/networking.md
```diff
@@ -0,0 +1,278 @@
...
+  - solidblock
+  - voleil
+  - daomephsta
+  - YTG123-Mods
+  - zulrang
+  - i509VCB
+  - nshak
+  - earthcomputer
+  - natanfudge
+  - modmuss50
```
its-miroma: added `parzivail` for the diagram, lowercased, alphabetized

```suggestion
authors:
  - daomephsta
  - dicedpixels
  - earthcomputer
  - flooferland
  - fxmorin
  - i509vcb
  - modmuss50
  - natanfudge
  - netuserget
  - nshak
  - parzivail
  - skycatminepokie
  - solidblock
  - voleil
  - wxffel
  - ytg123-mods
  - zulrang
```

develop/networking.md
```diff
@@ -0,0 +1,278 @@
...
+The code calls logic only present on the client distribution of the Minecraft. The reason for Mojang distributing the
+game in this way is to cut down on the size of the Minecraft Server JAR file. There isn't really a reason to include an
+entire rendering engine when your own machine will render the world.
+
+In a development environment, client-only classes are indicated by the `@Environment(EnvType.CLIENT)` annotation.
+
+### How Do We Fix the Crash? {#how-do-we-fix-the-crash}
+
+To fix this issue, you need to understand how Minecraft communicates between the game client and dedicated
+server.
```
its-miroma:
```suggestion
server:
```

its-miroma (changes requested):

develop/networking.md
```diff
@@ -0,0 +1,278 @@
...
+your game client can't directly access the server instance.
+
+## An Introduction to Networking {#an-introduction-to-networking}
+
+### Defining a Payload {#defining-a-payload}
+
+First, you need to define an `Identifier` used to identify our packet's payload. For this example our identifier will be
+`fabric-docs-reference:summon_lightning`.
+
+We can define this identifier in our **common initializer** for easy access.
```
Octol1ttle: It would be actually easier to put it in the payload class itself

develop/networking.md
```diff
@@ -0,0 +1,278 @@
...
+  @[code lang=java transclude={11-11}](@/reference/latest/src/main/java/com/example/docs/networking/basic/SummonLightningPayload.java)
+
+We have also overridden `getId` to return our payload ID.
+
+### Registering a Payload {#registering-a-payload}
+
+Before we send a packet with our custom payload, we need to register it.
+
+::: info
+`S2C` and `C2S` are two common suffixes that mean _Server-to-Client_ and _Client-to-Server_ respectively.
```
Octol1ttle, line 122: Given the presence of this note, we could be a good example for devs and rename the payloads to `SummonLightningS2CPayload` and `GiveGlowingEffectC2SPayload` respectively?

develop/networking.md
```diff
@@ -0,0 +1,278 @@
...
+Finally, we get all the players in the server world through `PlayerLookup` and send a packet to each player.
+
+@[code lang=java transclude={28-30}](@/reference/latest/src/main/java/com/example/docs/networking/basic/LightningTaterItem.java)
+
+::: info
+Fabric API provides `PlayerLookup`, a collection of helper functions that will look up players in a server.
+
+A term frequently used to describe the functionality of these methods is "_tracking_". It means that an entity or a
+chunk
+on the server is known to a player's client (within in view distance) and the entity or block entity should notify
```
Octol1ttle: "within their view distance" or "in their view distance"

develop/networking.md
```diff
@@ -0,0 +1,278 @@
...
+tracking clients of changes.
+
+Tracking is an important concept for efficient networking, so that only the necessary players are notified of changes by
+sending packets.
+:::
+
+Remember to register the item in our **common initializer**.
+
+@[code lang=java transclude={24-25}](@/reference/latest/src/main/java/com/example/docs/networking/basic/FabricDocsReferenceNetworkingBasic.java)
```
Octol1ttle: Seems out of place and not relevant to this guide?

develop/networking.md
```diff
@@ -0,0 +1,278 @@
...
+
+Now, if you add this mod to a server and when a player uses our Lightning Tater item, every player will see lightning
+striking at the user's position.
+
+<VideoPlayer src="/assets/develop/networking/summon-lightning.webm" title="Summon lightning using Lightning Tater" />
+
+### Sending a Packet to the Server {#sending-a-packet-to-the-server}
+
+Just like sending a packet to the client, we start by creating a custom payload. This time, when a player uses a
+Poisonous Potato on a living entity, we will apply the Glowing effect to it.
```
Octol1ttle: A bit nitpicky and rather subjective: I think the concept of server authority needs to be ingrained into the readers. What we're doing really here is "asking the server to apply the Glowing effect to it" or "requesting that the server applies the Glowing effect"

develop/networking.md
```diff
@@ -0,0 +1,278 @@
...
+
+@[code lang=java transcludeWith=:::give_glowing_effect_payload](@/reference/latest/src/main/java/com/example/docs/networking/basic/GiveGlowingEffectPayload.java)
+
+We pass in the appropriate codec along with a method reference to get the value from the Record to build this codec.
+
+`GIVE_GLOWING_EFFECT_PAYLOAD_ID` is defined in the **common initializer** for easy access.
```
Octol1ttle: ditto

reference/latest/src/client/java/com/example/docs/network/basic/FabricDocsReferenceNetworkingBasicClient.java
```diff
@@ -0,0 +1,66 @@
...
+import com.example.docs.networking.basic.SummonLightningPayload;
+
+public class FabricDocsReferenceNetworkingBasicClient implements ClientModInitializer {
+	@Override
+	public void onInitializeClient() {
+		// :::client_global_receiver
+		ClientPlayNetworking.registerGlobalReceiver(SummonLightningPayload.ID, (payload, context) -> {
+			ClientWorld world = context.client().world;
+
+			if (world != null) {
```
Octol1ttle: Are there clear standards on nesting within the docs' example code? I would turn this into `if (world == null) return;`

reference/latest/src/client/java/com/example/docs/network/basic/FabricDocsReferenceNetworkingBasicClient.java
```diff
@@ -0,0 +1,66 @@
...
+				BlockPos entityPos = payload.pos();
```
Octol1ttle: I would use `lightningPos`, `entityPos` sounds like it was acquired from `entity`

reference/latest/src/client/java/com/example/docs/network/basic/FabricDocsReferenceNetworkingBasicClient.java
```diff
@@ -0,0 +1,66 @@
...
+					entity.setPosition(entityPos.getX(), entityPos.getY(), entityPos.getZ());
+					world.addEntity(entity);
+				}
+			}
+		});
+		// :::client_global_receiver
+
+		// :::use_item_callback
+		UseItemCallback.EVENT.register((playerEntity, world, hand) -> {
+			if (world.isClient()) {
```
Octol1ttle: ditto (nesting)

reference/latest/src/client/java/com/example/docs/network/basic/FabricDocsReferenceNetworkingBasicClient.java
```diff
@@ -0,0 +1,66 @@
...
+				ItemStack usedItemStack = playerEntity.getStackInHand(hand);
+
+				if (usedItemStack.isOf(Items.POISONOUS_POTATO) && hand == Hand.MAIN_HAND) {
+					HitResult target = MinecraftClient.getInstance().crosshairTarget;
+
+					if (target != null && target.getType() == HitResult.Type.ENTITY) {
+						Entity targettedEntity = ((EntityHitResult) target).getEntity();
```
Octol1ttle: `targeted`
https://english.stackexchange.com/questions/205815/is-targetted-a-standard-british-english-spelling

reference/latest/src/client/java/com/example/docs/network/basic/FabricDocsReferenceNetworkingBasicClient.java
```diff
@@ -0,0 +1,66 @@
...
+
+		// :::use_item_callback
+		UseItemCallback.EVENT.register((playerEntity, world, hand) -> {
+			if (world.isClient()) {
+				ItemStack usedItemStack = playerEntity.getStackInHand(hand);
+
+				if (usedItemStack.isOf(Items.POISONOUS_POTATO) && hand == Hand.MAIN_HAND) {
+					HitResult target = MinecraftClient.getInstance().crosshairTarget;
+
+					if (target != null && target.getType() == HitResult.Type.ENTITY) {
```
Octol1ttle: Given that we're using `crosshairTarget` anyway, wouldn't it make more sense to use `UseEntityCallback` for this?

reference/latest/src/main/java/com/example/docs/networking/basic/FabricDocsReferenceNetworkingBasic.java
```diff
@@ -0,0 +1,41 @@
...
+
+	public void onInitialize() {
+		PayloadTypeRegistry.playS2C().register(SummonLightningPayload.ID, SummonLightningPayload.CODEC);
+		PayloadTypeRegistry.playC2S().register(GiveGlowingEffectPayload.ID, GiveGlowingEffectPayload.CODEC);
+
+		// :::server_global_receiver
+		ServerPlayNetworking.registerGlobalReceiver(GiveGlowingEffectPayload.ID, (payload, context) -> {
+			Entity entity = context.player().getWorld().getEntityById(payload.entityId());
+
+			if (entity != null && entity.isInRange(context.player(), 5) && entity instanceof LivingEntity livingEntity) {
```
Octol1ttle:
```suggestion
			if (entity instanceof LivingEntity livingEntity && entity.isInRange(context.player(), 5)) {
```
`instanceof` can't return `true` if the LHS is `null`

Octol1ttle (changes requested):

develop/networking.md
dicedpixels: Yep. Makes sense to keep them closer to where it's used. Changed.

develop/networking.md
dicedpixels: Yes. Let's keep the focus on networking. There's a whole guide if someone needs clarification about registering items.

reference/latest/src/client/java/com/example/docs/network/basic/FabricDocsReferenceNetworkingBasicClient.java
dicedpixels: I don't think so. But I agree with having less nesting.

reference/latest/src/client/java/com/example/docs/network/basic/FabricDocsReferenceNetworkingBasicClient.java
dicedpixels: Yep, seems `UseEntityCallback` is simpler in this case. I've changed to it.

reference/latest/src/client/java/com/example/docs/network/basic/FabricDocsReferenceNetworkingBasicClient.java
its-miroma: no, not yet

develop/networking.md
```diff
@@ -0,0 +1,266 @@
...
+
+::: warning
+Below code is for demonstration purposes only.
+:::
+
+Say you had a Wand which highlights the block you're looking, which will be visible to all
+nearby players:
+
+```java
+public class HighlightingWandItem extends Item {
```
IMB11, line 39: Needs to be in the reference mod in some way because of the `use` method, this will most certainly change in the future.

IMB11 (dismissed):
Very good quality page! Well done. Just a very minor nitpick and then LGTM I think!

develop/networking.md
dicedpixels: Given this is just a "bad example", I'd rather have it here than in the mod. I'd expect it'll be a while before Items would change.

However, I'm open to suggestions.

Octol1ttle: Maybe we could re-use a vanilla item again (blaze rod?), like we did with the Poisonous Potato

dicedpixels: Yes, I'm just reluctant to add any code that would crash the mod 😅.

its-miroma: That's exactly the reason why we need the code in the mod

dicedpixels: Then I'll add a simpler example like adding a particle. Block highlighting sounds complex to implement. I'm open to other bad examples as well.

dicedpixels: Now that I'm looking into this, how am I supposed to call a client class on the server when we're using split sources? I'd say we leave the example and perhaps add an info box highlighting how this wouldn't happen on a split source project.

modmuss50: You dont, as that would crash the game, split sources is correctly stopping you from doing it. With packets you would register the client side handler in the client source set.

dicedpixels: Yep, we can't demonstrate the bad example. I don't think it's necessary to include it in the reference mod. Even if the item's `use` method changes, it would be just a small piece of code to update.

its-miroma (dismissed):

develop/networking.md
```diff
@@ -0,0 +1,266 @@
...
+The code calls logic only present on the client distribution of the Minecraft. The reason for Mojang distributing the
+game in this way is to cut down on the size of the Minecraft Server JAR file. There isn't really a reason to include an
+entire rendering engine when your own machine will render the world.
+
+In a development environment, client-only classes are indicated by the `@Environment(EnvType.CLIENT)` annotation.
+
+### How Do We Fix the Crash? {#how-do-we-fix-the-crash}
+
+To fix this issue, you need to understand how Minecraft communicates between the game client and dedicated
+server:
```
Octol1ttle:
```suggestion
To fix this issue, you need to understand how the game client and dedicated
server communicate:
```

develop/networking.md
```diff
@@ -0,0 +1,266 @@
...
+| Connection Type               | Access to Game Client      |
+|-------------------------------|----------------------------|
+| Connected to dedicated Server | None (Server crash)        |
+| Connected over LAN            | Yes (Not host game client) |
+| singleplayer (or LAN host)    | Yes (Full access)          |
+
+It may seem complicated to have communication with the server in three different ways. However, you don't need to
+communicate in three different ways with the game client. Since all three connection types communicate with the game
+client using packets, you only need to communicate with the game client like you're always running on a dedicated
+server.
```
Octol1ttle:
```suggestion
```
The relation between "connection type" and "access to game client" is logically incorrect, since it is the client that connects to the server.
Access to the game client depends on the type of server: integrated or dedicated, but in the context of networking it doesn't matter

develop/networking.md
```diff
@@ -0,0 +1,266 @@
...
+Let's examine the code above.
+
+We can access the data from our payload by calling the Record's getter methods. In this case `payload.pos()`. Which then
+can be used to get the `x`, `y` and `z` positions.
+
+@[code lang=java transclude={37-37}](@/reference/latest/src/client/java/com/example/docs/network/basic/FabricDocsReferenceNetworkingBasicClient.java)
+
+Finally, we create a `LightningEntity` and add it to the world.
+
+@[code lang=java transclude={38-43}](@/reference/latest/src/client/java/com/example/docs/network/basic/FabricDocsReferenceNetworkingBasicClient.java)
```
Octol1ttle: image

develop/networking.md
```diff
@@ -0,0 +1,266 @@
...
+We create and instance of our `GiveGlowingEffectC2SPayload` with the necessary arguments. In this case, the network ID
+of
+the targeted entity.
+
+@[code lang=java transclude={56-56}](@/reference/latest/src/client/java/com/example/docs/network/basic/FabricDocsReferenceNetworkingBasicClient.java)
+
+Finally, we send a packet to the server by calling `ClientPlayNetworking.send` with the instance of our
+`UsePoisonousPotatoPayload`.
+
+@[code lang=java transclude={57-57}](@/reference/latest/src/client/java/com/example/docs/network/basic/FabricDocsReferenceNetworkingBasicClient.java)
```
Octol1ttle: image

develop/networking.md
```diff
@@ -0,0 +1,266 @@
...
+To send a packet, let's add an action when the player uses a Poisonous Potato. We'll be using the `UseEntityCallback`
+event to
+keep things concise.
+
+We register the event in our **client initializer**, and we use `isClient()` to ensure that the action is only triggered
+on the logical client.
+
+@[code lang=java transcludeWith=:::use_entity_callback](@/reference/latest/src/client/java/com/example/docs/network/basic/FabricDocsReferenceNetworkingBasicClient.java)
+
+We create and instance of our `GiveGlowingEffectC2SPayload` with the necessary arguments. In this case, the network ID
```
Octol1ttle:
```suggestion
We create an instance of our `GiveGlowingEffectC2SPayload` with the necessary arguments. In this case, the network ID
```

Octol1ttle (changes requested):

develop/networking.md
```diff
@@ -0,0 +1,252 @@
...
+
+- A public static instance of `CustomPayload.Id` to uniquely identify this custom payload. We will be referencing this
+  ID in both our common and client code.
+
+@[code lang=java transclude={14-14}](@/reference/latest/src/main/java/com/example/docs/networking/basic/SummonLightningS2CPayload.java)
+
+- A public static instance of a `PacketCodec` so that the game knows how to serialize/deserialize the contents of the
+  packet.
+
+@[code lang=java transclude={15-15}](@/reference/latest/src/main/java/com/example/docs/networking/basic/SummonLightningS2CPayload.java)
```
Octol1ttle, line 111: image

develop/networking.md
```diff
@@ -0,0 +1,252 @@
...
+
+### Registering a Payload {#registering-a-payload}
+
+Before we send a packet with our custom payload, we need to register it.
+
+::: info
+`S2C` and `C2S` are two common suffixes that mean _Server-to-Client_ and _Client-to-Server_ respectively.
+:::
+
+This can be done in our **common initializer** by `PayloadTypeRegistry.playS2C().register` which takes in a
```
Octol1ttle:
```suggestion
This can be done in our **common initializer** by using `PayloadTypeRegistry.playS2C().register`, which takes in a
```

develop/networking.md
```diff
@@ -0,0 +1,252 @@
...
+@[code lang=java transclude={25-25}](@/reference/latest/src/main/java/com/example/docs/networking/basic/FabricDocsReferenceNetworkingBasic.java)
+
+A similar method exists to register client-to-server payloads: `PayloadTypeRegistry.playC2S().register`.
+
+### Sending a Packet to the Client {#sending-a-packet-to-the-client}
+
+To send a packet with our custom payload, we can use `ServerPlayNetworking.send` which takes in a `ServerPlayerEntity`
+and a `CustomPayload`.
+
+Let's start by creating our Lightning Tater item. You can override `use` to add trigger an action when the item is used.
```
Octol1ttle:
```suggestion
Let's start by creating our Lightning Tater item. You can override `use` to trigger an action when the item is used.
```

develop/networking.md
```diff
@@ -0,0 +1,252 @@
...
+In this case, let's send packets to all the players in the server world.
```
Octol1ttle:
```suggestion
In this case, let's send packets to all players in the server world.
```

develop/networking.md
```diff
@@ -0,0 +1,252 @@
...
+
+We only send packets when the action is initiated on the server, by returning early with a `isClient` check:
+
+@[code lang=java transclude={22-24}](@/reference/latest/src/main/java/com/example/docs/networking/basic/LightningTaterItem.java)
+
+We create an instance of the payload with the user's position:
+
+@[code lang=java transclude={26-26}](@/reference/latest/src/main/java/com/example/docs/networking/basic/LightningTaterItem.java)
+
+Finally, we get all the players in the server world through `PlayerLookup` and send a packet to each player.
```
Octol1ttle:
```suggestion
Finally, we get all players in the server world through `PlayerLookup` and send a packet to each player.
```

develop/networking.md
```diff
@@ -0,0 +1,252 @@
...
+@[code lang=java transcludeWith=:::use_entity_callback](@/reference/latest/src/client/java/com/example/docs/network/basic/FabricDocsReferenceNetworkingBasicClient.java)
+
+We create an instance of our `GiveGlowingEffectC2SPayload` with the necessary arguments. In this case, the network ID
+of
+the targeted entity.
+
+@[code lang=java transclude={50-50}](@/reference/latest/src/client/java/com/example/docs/network/basic/FabricDocsReferenceNetworkingBasicClient.java)
+
+Finally, we send a packet to the server by calling `ClientPlayNetworking.send` with the instance of our
+`UsePoisonousPotatoPayload`.
```
Octol1ttle:
```suggestion
`GiveGlowingEffectC2SPayload`.
```

reference/latest/src/client/java/com/example/docs/network/basic/FabricDocsReferenceNetworkingBasicClient.java
```diff
@@ -0,0 +1,60 @@
...
+
+		// :::use_entity_callback
+		UseEntityCallback.EVENT.register((player, world, hand, entity, hitResult) -> {
+			if (!world.isClient()) {
+				return ActionResult.PASS;
+			}
+
+			ItemStack usedItemStack = player.getStackInHand(hand);
+
+			if (usedItemStack.isOf(Items.POISONOUS_POTATO) && hand == Hand.MAIN_HAND) {
```
Octol1ttle: The client may send a packet that'll be rejected by our check if we interact with a non-living entity (e.g. armor stand or minecart)
```suggestion
			if (entity instanceof LivingEntity && usedItemStack.isOf(Items.POISONOUS_POTATO) && hand == Hand.MAIN_HAND) {
```

Octol1ttle (changes requested):

reference/latest/src/client/java/com/example/docs/network/basic/FabricDocsReferenceNetworkingBasicClient.java
dicedpixels: Armor stands are living entities, but this would be valid for things like mine carts. So lets keep it.

develop/networking.md
```diff
@@ -0,0 +1,252 @@
...
+  - natanfudge
+  - netuserget
+  - nshak
+  - parzivail
+  - skycatminepokie
+  - solidblock
+  - voleil
+  - wxffel
+  - ytg123-mods
+  - zulrang
```
its-miroma:
```suggestion
authors:
  - Daomephsta
  - dicedpixels
  - Earthcomputer
  - FlooferLand
  - FxMorin
  - i509VCB
  - modmuss50
  - natanfudge
  - NetUserGet
  - NShak
  - parzivail
  - skycatminepokie
  - SolidBlock-cn
  - Voleil
  - Wxffel
  - YTG123-Mods
  - zulrang
```

develop/networking.md
dicedpixels: My bad, I have IJ configured to sort fields 😅. Fixed.

its-miroma: out of curiosity, how does your IDE sort them? alphabetically?

Octol1ttle (approved):

its-miroma (changes requested):

its-miroma:
still not convinced with the diagram being an image, but there's not much to do about it

its-miroma (approved):

develop/networking.md
```diff
@@ -0,0 +1,252 @@
...
+In a development environment, client-only classes are indicated by the `@Environment(EnvType.CLIENT)` annotation.
+
+### How Do We Fix the Crash? {#how-do-we-fix-the-crash}
+
+To fix this issue, you need to understand how the game client and dedicated server communicate:
+
+![Sides](/assets/develop/networking/sides.png)
+
+The diagram above shows that the game client and dedicated server are separate systems, bridged together using
+_packets_. Packets can contain data which we refer to as the _payload_.
```
modmuss50, line 79: This is fine I guess, it just doenst have anything to do with networking. Packets are still sent between the logicial client and server even in single player, or when on a LAN world. A better example would be to cause desync by setting something only the logicial client. Maybe try adding a block to the hotbar from a keybind or something.

PR #250, opened 2025-01-03
Update to 1.21.4: "Creating Your First Item"

raphm4044:
Since 1.21.4, Minecraft doesn't try to guess your model's location anymore, and so, we should now add an item description JSON in `assets/modid/items`, so this pull request aims to point out that in the documentation.

I only added the description for the Suspicious Substance, and not for the other items of the reference mod.

### DIFF
--- a/reference/1.20.4/src/main/resources/assets/fabric-docs-reference/items/suspicious_substance.json
+++ b/reference/1.20.4/src/main/resources/assets/fabric-docs-reference/items/suspicious_substance.json
@@ -1,0 +1,6 @@
+{
+  "model": {
+    "type": "minecraft:model",
+    "model": "fabric-docs-reference:item/suspicious_substance"
+  }
+}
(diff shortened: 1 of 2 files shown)


reference/1.20.4/src/main/resources/assets/fabric-docs-reference/items/suspicious_substance.json
IMB11, line 1: This has been created in the wrong folder (`1.20.4` instead of `latest`), we should also be data generating this instead of just placing raw JSON files.

raphm4044: Oops

IMB11:
Going to close this, tomorrow I will be working on the entire items section. I will definitely use what you've written about item descriptions and credit you. Thanks

raphm4044:
You're welcome!

PR #259, opened 2025-01-15
Exclude `translated` and `versions` from search

### LINKED ISSUE #256: Localized pages are not excluded from search
In english, type "registry" in the quick search. Scroll down and you'll find a dutch page
image

its-miroma:
Fix #256


IMB11 (approved):

IMB11:
Mod issues will be fixed by #258

.vitepress/config.mts
```diff
@@ -49,6 +49,14 @@ export default defineVersionedConfig(
 
     themeConfig: {
       search: {
+        options: {
```
IMB11, line 52: This should be done using the `exclude` option instead.

https://github.com/vuejs/vitepress/pull/2602

IMB11 (changes requested):

IMB11 (approved):

IMB11:
It seems like this has hit two birds with one stone, performance seems to be much much better with this in place.

PR #270, opened 2025-01-22
So many updates

### LINKED ISSUE #272: missing codeblock in Block States
https://docs.fabricmc.net/develop/blocks/blockstates#visualizing-the-property

Image

its-miroma:
> [!TIP]
> This diff page may have better performance than the review: <https://github.com/FabricMC/fabric-docs/compare/main...its-miroma:fmc:framework>

- Update `.github/CODEOWNERS`
- Update `.github/workflows`:
  - `build.yml`:
    - run on `main` only, otherwise pull requests from `l10n/main` run build twice
    - add conditions to avoid "Run failed" mails
  - `markdownlint.yml`:
    - remove `branches`, as `paths` already excludes pull requests from `l10n/main`
    - remove `workflow_dispatch`
  - `publish.yml`:
    - run on `ubuntu-24.04`
    - add conditions to avoid "Run failed" mails
- `settings.gradle`: build `latest` only
  - `update.ts`: don't add new versions to `settings.gradle`
- Add an `h1` heading from `$frontmatter.title` automatically
  - ~~remove all `# ` headings~~
  - `.markdownlint-cli2.yaml`: enable `MD025`
- add new `markdownlint` rules:
  - `max-one-sentence-per-line` (not in here)
  - `incorrect-mod-id`
  - `no-horizontal-rules`
  - `no-html-line-breaks`
  - `no-trailing-punctuation-in-lists
- enable lazy image loading
- add a footer
- format ~~many~~ few files with prettier
- add more vscode extension recommendations
- hide files that shouldn't be edited in vscode
- fix #272
- ~~fix formatting mistakes in `translated`~~
- ~~lint _all_ markdown files~~
- ~~delete `contributing.md` under `versions`~~


modmuss50 (dismissed):

.github/workflows/build.yml
its-miroma, line 1: add this job to avoid weird emails:

```yaml
      - name: Cleanup
        run: exit 0
```

.github/workflows/publish.yml
its-miroma, line 1: add this job to avoid weird emails:

```yaml
      - name: Cleanup
        run: exit 0
```

.github/workflows/publish.yml
```diff
@@ -32,6 +33,7 @@ jobs:
           path: "./.vitepress/dist"
 
   deploy:
```
its-miroma: This step is running on `ubuntu-22.04`, is this intentional?

modmuss50: No, this can be updated.

IMB11 (dismissed):

.markdownlint-cli2.yaml
```diff
@@ -122,11 +139,17 @@ config:
...
+        searchPattern: "/^( *- .*)[^a-z0-9?!`\"\\)]$/gi"
+        replace: "$1"
+        searchScope: text
       - name: no-trailing-whitespace
         message: "Don't use trailing whitespace"
         searchPattern: "/ +$/gm"
         replace: ""
         searchScope: code
 customRules:
+  # TODO: add      markdownlint-rule-max-one-sentence-per-line/rule.js
```
its-miroma, line 163: this would be helpful for a few reasons:

- no longer worrying about line wraps
- longer/complicated sentences stick out
- crowdin doesn't show a confusing new line in the middle of source strings

it would however lead to a monstrous diff, so i avoided doing it here.

if i ever wanted to add it, the best way would be to render the markdown files after the sentences split and run a `diff` on the built files to check for broken/changed formatting

.vitepress/config.mts
```diff
@@ -30,18 +30,21 @@ export default defineVersionedConfig(
         // Use the snippet plugin (transclusion, etc.)
         md.use(snippetPlugin);
       },
+      image: {
+        lazyLoading: true,
+      },
```
its-miroma, line 35: would appreciate feedback/testing on this

.vitepress/i18n.ts
```diff
@@ -220,6 +220,14 @@ function generateTranslatedThemeConfig(localeCode: string): Fabric.ThemeConfig {
 
     externalLinkIcon: true,
 
+    footer: {
+      copyright: websiteResolver("footer.copyright").replace(
+        "%s",
+        `<a href=\"https://github.com/FabricMC/fabric-docs/blob/main/LICENSE\" target=\"_blank\" rel=\"noreferrer\">${websiteResolver("footer.license")}</a>`
```
its-miroma: unsure what the best way to sanitize `footer.license` is

a malicious translation could close the `</a>`...

.vitepress/theme/index.ts
```diff
@@ -16,26 +16,33 @@ import "./style.css";
...
   },
   Layout() {
+    const { page, frontmatter } = useData();
+
     const children = {
-      "doc-before": () => h(VersionReminder),
+      "doc-before": () => [
+        frontmatter.value.title
+          ? h("h1", { class: "vp-doc" }, frontmatter.value.title)
+          : null,
```
its-miroma, line 38: is there ever a need for two different titles in `<head>` and `<body>`?

.vitepress/theme/style.css
```diff
@@ -8,7 +8,7 @@
   font-family: "Flag Emoji Polyfill";
   unicode-range: U+1F1E6-1F1FF, U+1F3F4, U+E0062-E0063, U+E0065, U+E0067,
     U+E006C, U+E006E, U+E0073-E0074, U+E0077, U+E007F;
-  src: url("./CountryFlagsPolyfill.woff2") format('woff2');
+  src: url("./CountryFlagsPolyfill.woff2") format("woff2");
```
its-miroma, line 11: if we used Noto Color Emoji we could embed it without shipping the binary file in the repo

contributing.md
```diff
@@ -3,8 +3,6 @@ title: Contribution Guidelines
 description: Guidelines for contributions to the Fabric Documentation.
 ---
 
-# Contribution Guidelines {#contributing}
-
```
its-miroma: every single markdown file will have at least a two-line diff.

I've also deleted some `contributing` files which slipped through in `versions`

its-miroma (commented):
~~this PR may be unreviewable~~

.vitepress/theme/style.css
IMB11: We dont need to include all emojis though, we only need the flags.

.vitepress/theme/index.ts
IMB11: Yes, the homepage

its-miroma: `doc-before` (line 35) only includes doc pages, not the homepage nor the 404 page

.github/workflows/build.yml
```diff
@@ -37,3 +43,8 @@ jobs:
         with:
           reports: |
             reference/**/build/reports/checkstyle/*.xml
+
+  cleanup:
+    runs-on: ubuntu-24.04
+    steps:
+      - run: exit 0
```
modmuss50: Why is this needed?

.github/CODEOWNERS
```diff
@@ -1,8 +1,18 @@
 * @FabricMC/documentation
 
+.gitattributes @FabricMC/developers
```
modmuss50: I think we should changed this file to default to devs, but allow the docs team access to the content.

.github/workflows/build.yml
```diff
@@ -1,12 +1,17 @@
 name: Build
-on: [push, pull_request]
+
+on:
+  push:
+    branches: ["main"]
```
modmuss50: Why would we only want to build the main branch?

.github/workflows/build.yml
```diff
@@ -1,12 +1,17 @@
...
+  pull_request:
 
 concurrency:
   group: build-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
 jobs:
   docs:
+    if: github.repository == 'FabricMC/fabric-docs'
```
modmuss50: This doesnt make any sense for building, if you fork the repo the actions should still run. Only the publish task needs this.

.github/workflows/publish.yml
```diff
@@ -16,6 +16,7 @@ permissions:
 
 jobs:
   build:
+    if: github.repository == 'FabricMC/fabric-docs'
```
modmuss50: Only needed for the deploy job.

.markdownlint-cli2.yaml
```diff
@@ -9,16 +11,15 @@ config:
...
     siblings_only: true
-  MD025:
-    front_matter_title: ""
   MD026:
     # Allow headings to end with `!`
     punctuation: ".,;:。，；："
   MD029:
     style: ordered
-  MD033: false
+  MD033: false # TODO: remove
```
modmuss50: TODO? Is this for the future?

.vitepress/config.mts
modmuss50: Id argue we dont need this? By the time you scroll down the image it will have already been loaded.

.vitepress/i18n.ts
modmuss50: Maybe use `document.createElement("a")` with ``.textContent and then get the `outerHTML` Im not sure if this works in this context though?

.github/workflows/build.yml
its-miroma: Because `pull_request` (line 6) already builds the docs for incoming contributions

.github/workflows/build.yml
its-miroma: See explanation following this message on Discord: <https://discord.com/channels/507304429255393322/1208846408552030238/1332670446818299944>

TL;DR: syncing a fork's main branch would trigger the job uselessly and cause it to fail and send annoying notifications.

.github/workflows/build.yml
its-miroma: See this message on Discord: <https://discord.com/channels/507304429255393322/1208846408552030238/1331740336640233584>

TL;DR: not running any job still sends annoying mail

.github/workflows/publish.yml
its-miroma: See explanation above

.markdownlint-cli2.yaml
its-miroma: That's the no HTML tags rule, it's possible to set an allowlist which may be safer, but I don't know if that's worth it

.vitepress/config.mts
its-miroma: Lazy loading of images allows faster loading of a page because images get loaded when needed.

Check out <https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading>

.github/workflows/build.yml
modmuss50: I disagree with it being useless, it should succeed. We dont do this with any other projects so I dont understand why its an issue here.

.github/workflows/build.yml
modmuss50: Again, this doesnt match what we have for other repos, so I dont get whats special here?

its-miroma: If this is done on other repos as well, then the same applies there. Some builds are redundant, and these are not any more special than the others'. If you don't believe it's worth fixing we can revert these changes.

modmuss50: I get why we wouldnt want to try to deploy from other branches, but it should still attempt to build. Can you disable actions on your branch if you dont want to waste builds (at Githubs expense not yours).

.github/workflows/build.yml
its-miroma: That's my point: it _will_ most certainly succeed because syncing a repo's `main` branch just means pulling the (successful) commits from upstream.

You can also consider it to be a security risk, because if it is ran on updates to `main` on forks, a malicious commit could edit the action, and the new action might get ran without any approval (I'd need to test it to confirm these suspicion).

modmuss50: > a malicious commit

The malicious commit would need to make it into our main branch, that should never happen. And could only happen if someone approved the PR.

.github/workflows/build.yml
its-miroma: Why would we want to build the docs and the mod on branches if a `pull_request` is not open?
Also, one could say that Netlify builds the docs, so that entire job is redundant.

I get why you'd be reluctant to change these actions, just confirm that I should undo these if needed

> Can you disable actions on your branch if you dont want to waste builds

I'm not sure how I can do that...

> at Githubs expense not yours

at FabricMC's minutes expense, I don't know how much of a problem that is

.github/workflows/build.yml
its-miroma: that's reassuring; that still leaves the first part relevant

modmuss50: I dont see a major problem with that, in this one edge case it will build something that is known to already build, seems like a very minor issue that isnt our problem.

modmuss50 (approved):

modmuss50:
Looks good, we can either merge this and then get the lint fixes, or I can just slap a re approval with the lint fixes as part of this. Thanks for splitting it out. In the future please try to keep PRs to just 1 or two changes.

its-miroma:
let's merge this PR and keep it clean, unless @IMB11 wants to review it.
The linting fixes are in #282

PR #280, opened 2025-02-12
Credits Page

its-miroma:
A simple Credits page with a list of GitHub maintainers and contributors, Crowdin translators, and page `authors`.

- [ ] `CROWDIN_TOKEN`
- [ ] `GITHUB_TOKEN`

| Name              | Link                                                                |
| :---------------: | :-----------------------------------------------------------------: |
| 😎 Deploy Preview | <https://deploy-preview-280--nimble-elf-d9d491.netlify.app/credits> |


.vitepress/data/maintainers.data.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export default {
+  async load() {
+    const members = await octokit.paginate(
+      octokit.rest.teams.listMembersInOrg,
+      {
+        org: "FabricMC",
+        team_slug: "documentation",
+      }
+    );
```
its-miroma:
```suggestion
    const members = (() => {
      try {
        return await octokit.paginate(
          octokit.rest.teams.listMembersInOrg,
          {
            org: "FabricMC",
            team_slug: "documentation",
          }
        );
      } catch (e) {
        // Allows build without a GITHUB_TOKEN
        return [];
      }
    })();
```

package.json
```diff
@@ -21,10 +22,13 @@
     "@rollup/rollup-linux-x64-gnu": "4.21.3"
   },
   "dependencies": {
+    "@crowdin/crowdin-api-client": "^1.41.1",
     "@vueuse/core": "^12.4.0",
+    "axios": "^1.7.9",
```
modmuss50: Can we not use `fetch`? I am really worrying about how many depedencies the docs site is starting to have.

.vitepress/data/committers.data.ts
```diff
@@ -0,0 +1,48 @@
...
+  async load() {
+    const contributors = await (async () => {
+      try {
+        return await octokit.paginate(octokit.rest.repos.listContributors, {
+          owner: "FabricMC",
+          repo: "fabric-docs",
+        });
+      } catch (e) {
+        // Allows build without internet
+        return [];
```
modmuss50: This isnt good as it will silently fail if there is an error making this request. It should bail out before any request is made if the github token isnt set.

.vitepress/data/committers.data.ts
```diff
@@ -0,0 +1,48 @@
...
+      }
+    })();
+    const { data } = await (async () => {
+      try {
+        return await octokit.rest.users.getByUsername({
+          username: "FabricMCBot",
+        });
+      } catch (e) {
+        // Allows build without internet
+        return { data: { avatar_url: "" } };
```
modmuss50: Same here.

.vitepress/data/maintainers.data.ts
```diff
@@ -0,0 +1,53 @@
...
+  async load() {
+    const members = await (async () => {
+      try {
+        return await octokit.paginate(octokit.rest.teams.listMembersInOrg, {
+          org: "FabricMC",
+          team_slug: "documentation",
+        });
+      } catch (e) {
+        // Allows build without a GITHUB_TOKEN or without internet
+        return [];
```
modmuss50: Same here, dont even try to make the request when the token isnt set.

.vitepress/data/maintainers.data.ts
```diff
@@ -0,0 +1,53 @@
...
+    return (
+      await Promise.all(
+        members.map(async (member) => ({
+          val: await (async () => {
+            try {
+              return await octokit.paginate(octokit.rest.orgs.listForUser, {
+                username: member.login,
+              });
+            } catch (e) {
+              // Allows build without internet
```
modmuss50: ditto

.vitepress/data/translators.data.ts
```diff
@@ -0,0 +1,108 @@
...
+    return new crowdin.default({ token: CROWDIN_TOKEN });
+  } catch (e) {
+    // Allows build without a CROWDIN_TOKEN or without internet
+    return { sourceFilesApi: {}, reportsApi: {} };
+  }
+})();
+
+export default {
+  async load() {
+    if (!CROWDIN_TOKEN) return [];
```
modmuss50, line 20: Yes, this is good, the others should do this with the github token.

.vitepress/data/translators.data.ts
```diff
@@ -0,0 +1,108 @@
...
+      }
+
+      report = (
+        await axios.get(
+          (
+            await reportsApi.downloadReport(PROJECT_ID, identifier)
+          ).data.url
+        )
+      ).data.data;
+    } catch (e) {}
```
modmuss50: Why catch here and do nothing if something goes wrong?

.vitepress/data/translators.data.ts
```diff
@@ -0,0 +1,108 @@
...
+      translators.set(user.username, user);
+    });
+
+    return Array.from(translators.values())
+      .map(
+        (contributor) =>
+          ({
+            avatar:
+              contributor.avatar ||
+              "https://i2.wp.com/crowdin.com/images/user-picture.png?ssl=1",
```
modmuss50: Question: Are the images loaded from crowdin or are they duplicated into our static build?

modmuss50 (commented):
Error handling needs fixing.

I do worry this docs site is becoming very complex now, I guess its fine.

.vitepress/data/translators.data.ts
its-miroma: this would error out if the profile is private, which only means we'd not get the avatar

package.json
its-miroma: i can't remember if I'd tried and it didn't work, will try again

.vitepress/data/translators.data.ts
its-miroma: i think both github and crowdin avatars are loaded, not downloaded

.vitepress/data/translators.data.ts
modmuss50: Can you catch the specific error in that case?

.vitepress/data/translators.data.ts
modmuss50: Ok, I think this is acceptable. It's not ideal using a hardcoded url like this that seems likely to change, the alternative solution is likely even more complex.

its-miroma: we could use FabricMCBot's profile picture like with GitHub

.vitepress/data/bot.data.ts
its-miroma, line 1: once #293 drops I'll update it to use the Avatater

.vitepress/data/translators.data.ts
its-miroma, line 1: this still needs testing as I don't have a Crowdin token

credits.md
```diff
@@ -0,0 +1,101 @@
...
+    <template #members>
+      <VPTeamMembers :members="authors" size="small" />
+    </template>
+  </VPTeamPageSection>
+  <VPTeamPageSection v-if="translators.length">
+    <template #title>{{ options!.translators.title }}</template>
+    <template #members>
+      <VPTeamMembers :members="translators" size="small" />
+    </template>
+  </VPTeamPageSection>
```
its-miroma, line 101: how do you prefer these to be sorted?

IMB11: I would probs split this up into github and crowdin

IMB11: Also this page is only accessible by URL

PR #285, opened 2025-02-14
delete redundant `h1` in translated files

its-miroma:


modmuss50:
Im not convinced we should be touching the translated files like this, my understanding was that crowdin was the source of truth. Im happy to merge it and see what happens.

modmuss50 (approved):
lets try this, worse case they all come back.

its-miroma:
I love your spirit

PR #294, opened 2025-02-27
Make HudRenderingEntrypoint less obtrusive

Earthcomputer:
Before:
Peek 2025-02-27 19-55

After:
Peek 2025-02-27 20-06

A variable that used to be in ticks was accidentally changed to milliseconds during the 1.21.4 update. I also made it smaller.


its-miroma (changes requested):
please update the gif as well

its-miroma (approved):

PR #293, opened 2025-02-27
Update Authors component

its-miroma:
- render it above content if the sidebar is hidden when window width is small
- ~~use a tater-themed profile picture as fallback~~ (thank you anyway @Alexander01998)
- sort authors all together (both those with and without github)

waiting for confirmation that **tiny_potato** is compatible with the LICENSE to push `/public/assets/avatater.png`

PR #298, opened 2025-03-06
refactor: use `computed` instead of `ref`

its-miroma:
AuthorsComponent is handled by #293


modmuss50 (dismissed):
I dont really know what im looking at, but I trust its better.

its-miroma:
it's what we're already using for `VersionSwitcher` I think

IMB11 (approved):
LGTM @FabricMC/developers

modmuss50 (approved):

Issue #230, opened 2024-12-15
Developer Guides: Sidedness


Android789515:
https://docs.fabricmc.net/develop/items/first-item#preparing-your-items-class This section here does not specify where to make the `ModItems` class.  Whether in main or client folder.

---

### Maintainer notes:

- [ ] Create new "sidedness" page that goes through client, server, integrated servers, etc. - A spiritual improvement of the wiki's sidedness page.

Linguardium:
Items would be common since both logical client (rendering) and logical server (logic) needs to know about them. The differences in client and main should have been covered in a previous section of the docs

Android789515:
It could be elaborated more.  It just says

The src/main/java folder is used to store the Java source code for your mod - it exists on both the client and server environments.

Linguardium:
Agreed. the logical server/logical client concept should be elaborated on the docs. it has not yet replaced the wiki page on it https://wiki.fabricmc.net/tutorial:side

JR1811:
I created a (very high level) overview on that some time ago, if someone wants to use it on a potential PR

![Image](https://github.com/user-attachments/assets/be9cc475-6e63-489d-94ae-ef7e92c75e6b)

cassiancc:
Prior art also includes https://github.com/natanfudge/fabric-docs/blob/master/docs/modding-tutorials/conventions-and-terminology/side.md

PR #299, opened 2025-03-07
Explain networking

modmuss50:
I struggled to properly explain this, I know its a bit short but I hope its better.


develop/networking.md
```diff
@@ -26,69 +26,28 @@ so this page is split up into a few categories.
 
 ## Why Is Networking Important? {#why-is-networking-important}
 
-The importance of networking can be shown by a simple code example.
+Packets of data are the core concept of networking in Minecraft. Packets are made up of arbitrary data that can be sent either from
```
Octol1ttle:
```suggestion
Packets are the core concept of networking in Minecraft. Packets are made up of arbitrary data that can be sent either from
```

develop/networking.md
```diff
@@ -26,69 +26,28 @@ so this page is split up into a few categories.
...
+the server to the client or from the client to the server. Almost everything you do in the game involves networking in some way, whether
```
Octol1ttle:
```suggestion
server to client or from client to server. Almost everything you do in the game involves networking in some way, whether
```

develop/networking.md
```diff
@@ -26,69 +26,28 @@ so this page is split up into a few categories.
...
+you know it or not. For example, when you jump in the game, a packet is sent to the server to tell it that you jumped. The server then
+sends a packet to all the other clients to tell them that you jumped.
```
Octol1ttle: I'm not sure how much it matters, but this doesn't sound factually correct 😅 AFAIK a jump action (and other forms of player movement) just gets boiled down to a change in position and velocity

develop/networking.md
```diff
@@ -26,69 +26,28 @@ so this page is split up into a few categories.
...
-[Server thread/FATAL]: Error executing task on Server
-java.lang.RuntimeException: Cannot load class net.minecraft.client.MinecraftClient in environment type SERVER
-```
-
-### Why Does the Server Crash? {#why-does-the-server-crash}
-
-The code calls logic only present on the client distribution of the Minecraft. The reason for Mojang distributing the
-game in this way is to cut down on the size of the Minecraft Server JAR file. There isn't really a reason to include an
-entire rendering engine when your own machine will render the world.
+One imporant thing to keep in mind is even in single player there is still a server running, Packets are still used to communicate between
```
Octol1ttle:
```suggestion
One imporant thing to keep in mind is there is always a server running, even in singleplayer and LAN. Packets are still used to communicate between
```

develop/networking.md
```diff
@@ -26,69 +26,28 @@ so this page is split up into a few categories.
...
+the client and server even when no one else is playing with you. We usually call this the logicial client/server or the client/server thread.
+This is diffrent from the client/server code split, that prevents you from being able to use client only classes in common (shared between
+the client and server) code.
```
Octol1ttle:
```suggestion
the client and server even when no one else is playing with you. When talking about sides in networking, the terms "**logical** client" and "**logical** server" are used. The integrated singleplayer/LAN server and the dedicated server are both **logical** servers, but only the dedicated server can be considered a **physical** server.
```

develop/networking.md
```diff
@@ -26,69 +26,28 @@ so this page is split up into a few categories.
...
-The diagram above shows that the game client and dedicated server are separate systems, bridged together using
-_packets_. Packets can contain data which we refer to as the _payload_.
-
-This packet bridge does not only exist between a game client and dedicated server, but also between your client and
-another client connected over LAN. The packet bridge is also present even in singleplayer. This is because the game
-client will spin up a special integrated server instance to run the game on.
-
-Connection to a server over LAN or singleplayer can also be treated like the server is a remote, dedicated server; so
-your game client can't directly access the server instance.
+When state is not synced between the client and server, you can run into issues where the server or other clients dont agree with what the
```
Octol1ttle:
```suggestion
When state is not synced between the client and server, you can run into issues where the server or other clients don't agree with what another
```

develop/networking.md
```diff
@@ -26,69 +26,28 @@ so this page is split up into a few categories.
...
+client is doing. This is often known as a "desync". When writing your own mod you may need to send a packet of data to notify the server,
+client or any other connected clients that something has happened.
```
Octol1ttle: It is the server's job to notify other clients about changes. Merging "notify server (from client)" and "notify client (from server)" in the same sentence like this is confusing

Octol1ttle (changes requested):

develop/networking.md
modmuss50: Thats true, ill change this to a chat message. Just wanted something simple to get the idea across.

develop/networking.md
modmuss50: Hopefully its better now.

develop/networking.md
```diff
@@ -26,69 +26,26 @@ so this page is split up into a few categories.
...
-[Server thread/FATAL]: Error executing task on Server
-java.lang.RuntimeException: Cannot load class net.minecraft.client.MinecraftClient in environment type SERVER
-```
-
-### Why Does the Server Crash? {#why-does-the-server-crash}
-
-The code calls logic only present on the client distribution of the Minecraft. The reason for Mojang distributing the
-game in this way is to cut down on the size of the Minecraft Server JAR file. There isn't really a reason to include an
-entire rendering engine when your own machine will render the world.
+One imporant thing to keep in mind is there is always a server running, even in singleplayer and LAN. Packets are still used to communicate between
```
Octol1ttle:
```suggestion
One important thing to keep in mind is there is always a server running, even in singleplayer and LAN. Packets are still used to communicate between
```

Octol1ttle (approved):

develop/networking.md
```diff
@@ -26,69 +26,26 @@ so this page is split up into a few categories.
 
 ## Why Is Networking Important? {#why-is-networking-important}
 
-The importance of networking can be shown by a simple code example.
+Packets are the core concept of networking in Minecraft. Packets are made up of arbitrary data that can be sent either from
+server to client or from client to server. Almost everything you do in the game involves networking in some way, whether
+you know it or not. For example, when you send a chat message, a packet is sent to the server with the content. The server then
+sends another packet to all the other clients with your message.
```
its-miroma: Either this:

```suggestion
Packets are the core concept of networking in Minecraft.
Packets are made up of arbitrary data that can be sent either from server to client or from client to server.
Check out the diagram below, which provides a visual representation of the networking architecture in Fabric:

![Sided Architecture](/assets/develop/networking/sides.png)

Notice how packets are the bridge between the server and the client; that's because almost everything you do in the game involves networking in some way, whether you know it or not.
For example, when you send a chat message, a packet is sent to the server with the content.
The server then sends another packet to all the other clients with your message.
```

...or delete [sides.png](https://github.com/FabricMC/fabric-docs/blob/daff2047d7cb7ebcb4c12e37164622449ce0c871/public/assets/develop/networking/sides.png) if deemed unnecessary.

<details>
<summary>Sided Architecture</summary>

Sided Architecture

</details>

its-miroma (changes requested):

IMB11:
This is good, just needs change requests resolved :)

IMB11:
Maybe worth using this diagram as well? https://github.com/FabricMC/fabric-docs/issues/230#issuecomment-2707025996

its-miroma (approved):

PR #304, opened 2025-03-11
Fix `git clone` command in russian documentation

bleudev:


its-miroma:
Thank you for your contribution.

Please make changes to translated files on [Crowdin](https://crowdin.com/project/fabricmc/ru)

its-miroma:
actually, this is not translatable on Crowdin, unsure what's going on...

could you try to fix this (outdated) string in particular: https://crowdin.com/editor/fabricmc/516/en-ru?view=comfortable&filter=basic&value=0#q=git
~~perhaps that might trigger a refresh~~

Edit: I just fixed some strings which may be enough to trigger said update

PR #331, opened 2025-05-25
Update to 1.21.5: Block Model Generation

Fellteros:
Updated the block model generation page to 1.21.5 and optimized imports.


its-miroma:
superseded by #347.

@Fellteros, would you want to take a look at the changes there?

Fellteros:
Yea, why not, though it should roughly the same as in 1.21.5 I think

PR #338, opened 2025-07-02
Running Game Tests

kevinthegreat1:
Adds instructions on how to run game tests locally and on github actions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,7 +38,7 @@ jobs:
           reports: |
             reference/**/build/reports/checkstyle/*.xml
 
-# :::automatic-testing:game-test:run-client-game-test
+# :::automatic-testing:game-test:3
   client_game_test:
     runs-on: ubuntu-24.04
     steps:
@@ -50,15 +50,15 @@ jobs:
           distribution: 'microsoft'
           java-version: '21'
 
-      - name: Run client gametest
-        run: ./gradlew runClientGametest
-# :::automatic-testing:game-test:run-client-game-test
+      - name: Run production client gametest
+        run: ./gradlew prodClient
+# :::automatic-testing:game-test:3
         working-directory: ./reference
-# :::automatic-testing:game-test:run-client-game-test
+# :::automatic-testing:game-test:3
       - name: Upload test screenshots
         uses: actions/upload-artifact@v4
         if: always()
         with:
           name: Test Screenshots
           path: build/run/clientGameTest/screenshots
-# :::automatic-testing:game-test:run-client-game-test
+# :::automatic-testing:game-test:3
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,12 +50,10 @@ jobs:
           distribution: 'microsoft'
           java-version: '21'
 
-      - name: Run client gametest with Xvfb
-        uses: modmuss50/xvfb-action@v1
-        with:
-          run: ./gradlew runClientGametest
+      - name: Run client gametest
+        run: ./gradlew runClientGametest
 # :::automatic-testing:game-test:run-client-game-test
-          working-directory: ./reference
+        working-directory: ./reference
 # :::automatic-testing:game-test:run-client-game-test
       - name: Upload test screenshots
         uses: actions/upload-artifact@v4
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,7 +51,7 @@ jobs:
           java-version: '21'
 
       - name: Run production client gametest
-        run: ./gradlew prodClient
+        run: ./gradlew runProductionClientGameTest
 # :::automatic-testing:game-test:3
         working-directory: ./reference
 # :::automatic-testing:game-test:3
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -37,3 +37,30 @@ jobs:
         with:
           reports: |
             reference/**/build/reports/checkstyle/*.xml
+
+# :::automatic-testing:game-test:run-client-game-test
+  client_game_test:
+    runs-on: ubuntu-latest
```
its-miroma: All software in Actions must be locked to a specific version.

```suggestion
    runs-on: ubuntu-24.04
```

.github/workflows/build.yml
```diff
@@ -37,3 +37,30 @@ jobs:
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Set up JDK 21
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'microsoft'
+          java-version: '21'
+
+      - name: Run client gametest with Xvfb
+        uses: modmuss50/xvfb-action@v1
```
its-miroma: I hope this is a trusted publisher, otherwise we'll have to tell modmuss to fork it and lock it to a commit SHA.

🐧

develop/automatic-testing.md
```diff
@@ -123,3 +123,13 @@ You can now create server and client game tests in the `src/gametest/java` direc
 :::
 
 See the respective Javadocs in Fabric API for more info.
+
+### Running Game Tests {#running-game-tests}
+
+Server game tests will be run automatically with the `build` Gradle task. You can run client game tests with the `runClientGameTest` Gradle task.
+
+### Setting Up GitHub Actions {#setting-up-github-actions-game-tests}
```
its-miroma: I think we need a `## Setting Up GitHub Actions` that includes:

- Action for building (or is the Mod Template enough?)
- Additional instructions for Unit Tests, from above
- Setup for Game Tests

If not, at least we should change the titles to avoid conflicts:

```suggestion
### Run Game Tests on GitHub Actions {#run-game-tests-on-github-actions}
```

.github/workflows/build.yml
modmuss50: This is no longer required, loom has builtin support for xvfb now.

kevinthegreat1: I tried removing it but I get this: https://github.com/FabricMC/fabric-docs/actions/runs/16023196874/job/45204804012

modmuss50: Ah, I see. You need to use a production run taks for this. See: https://docs.fabricmc.net/develop/loom/production-run-tasks#client

Let me know if you need help setting this up.

kevinthegreat1: Sorry for having so much trouble, but I'm running into another error: https://github.com/FabricMC/fabric-docs/actions/runs/16023720884/job/45206506705. Maybe it is best if you set it up. I have done what looks correct to me.

modmuss50: You need to also put fabric API on the `productionRuntimeMods` configuration, as its tottaly seperate from the regular configs.

See:
https://github.com/FabricMC/fabric-loom/blob/fdadd0575d6381e8171a44fd1cdf148fcacf9ae9/src/test/groovy/net/fabricmc/loom/test/integration/RunConfigTest.groovy#L183

kevinthegreat1:
The test seems to have gotten stuck...

its-miroma:
Should I cancel?

.github/workflows/build.yml
```diff
@@ -37,3 +37,28 @@ jobs:
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Set up JDK 21
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'microsoft'
+          java-version: '21'
+
+      - name: Run production client gametest
```
its-miroma, line 53: Let's try running the action again. Is 10 minutes a good max time?

```suggestion
      - name: Run production client gametest
        timeout-minutes: 10
```

kevinthegreat1:
Turns out I literally didn't even add gametest to the jvm arg... Ooops.

This pr now works as expected. The only problem is that I couldn't get code group to work with the enhanced snippets, so the last two snippets are two snippets instead of one code group, which shouldn't be a big issue.

its-miroma (dismissed):
LGTM!

I don't think we can use code groups with transclusions.

reference/latest/build.gradle
```diff
@@ -23,3 +29,20 @@ fabricApi {
...
+
+// :::automatic-testing:game-test:2
+tasks.register("runProductionClientGameTest", net.fabricmc.loom.task.prod.ClientProductionRunTask) {
+	jvmArgs.addAll([
+			"-Dfabric.client.gametest",
+			"-Dfabric-tag-conventions-v2.missingTagTranslationWarning=fail",
+			"-Dfabric-tag-conventions-v1.legacyTagWarning=fail",
+			"-enableassertions",
+			"-Dmixin.debug.verify=true",
+			"-Dmixin.debug.countInjections=true",
```
modmuss50: I dont think you need any of thsese.

kevinthegreat1: Done in 362a533.

modmuss50 (approved):

its-miroma (approved):

PR #339, opened 2025-07-31
Create confuse.md

reuben3ym:


dicedpixels:
@reuben3ym Why is this necessary?

cputnam-a11y:
Could you explain how this contribution is intended to improve the docs?

cputnam-a11y (changes requested):
Generally seems pointless

reuben3ym:
@dicedpixels too much differences in the docs,for example :
example code "https://github.com/FabricMC/fabric-docs/blob/main/reference/latest/src/client/java/com/example/docs/rendering/blockentity/CounterBlockEntityRenderer.java" is exactly right, but  the code in "https://docs.fabricmc.net/develop/blocks/block-entity-renderer" lack a lot.

dicedpixels:
Instead of providing constructive criticism, you decided to make this useless PR?

Don't complain about something provided to you at 0 cost. You're not entitled to anything.

reuben3ym:
If you choose to do it, please try your best to do it well

PR #350, opened 2025-08-20
Build by version

kevinthegreat1:
Adds a build script that builds the docs version by version and update workflows to use `build_by_version`.

`build_by_version` can run with default node RAM, and I've tested the two build options on my laptop (m1 macbook air).
Build vs build by version memory usage

[Generates identical html on the main branch*](https://github.com/kevinthegreat1/fabric-docs/compare/5d23bb19..a0a43d88)

_\* I didn't actually check all the files. I only checked a couple._


kevinthegreat1:
Unfortunately, the preview seems to be janky with old versions. It gives a 404 but works once you reload the page.

kevinthegreat1:
Unfortunately, memory usage seems to have gone up due to more versions and pages. This does not appear to be as scalable as imagined, but it's definitely still better than what we have currently.

its-miroma:
Closing because this is no longer as necessary. Could be reopened down the line.

PR #348, opened 2025-08-19
Clarify that `mod-id` placeholders must be customized

fatherOfMS:
mod-id should be replace with specific mod-id, and shoud notice beginers, or will confusing them for lots time and i am the unlucky poor guy stucke for 2 hours!!!!

What was written earlier is too misleading.

PR #352, opened 2025-08-21
notice user replace mod-id in doument with their specific mod's id

fatherOfMS:
notice user replace mod-id in doument with their specific mod's id


its-miroma:
Thank you. Let's keep working in #348.

Duplicate of #348

PR #355, opened 2025-08-27
[CN] added missing translation

sp2dev:


its-miroma:
Hello @sp2dev, thank you for your contribution.

However, we do not accept localization fixes through GitHub; please use [Crowdin](https://crowdin.com/project/fabricmc/zh-CN).

These untranslatable strings are a known issue, and we're working to fix that.

PR #357, opened 2025-09-09
Change Mod ID to `ExampleMod`

### LINKED ISSUE #162: Use `ExampleMod` instead of `FabricDocsReference`
Mentioned in Discord as well - I think it is worth examining whether the docs would be better off using names like `example-mod` and `ExampleModInitializer` in the example code rather than `fabric-docs-reference` and `FabricDocsReference`. I notice in the mod dev channels, that it seems to be a pretty common point of confusion. Many newcomers have been asking questions such as "what is FabricDocsReference?" and "where do I find FabricDocsReference?" because it is not immediately clear that these classes and names should be replaced with THEIR mod's ModInitializer class. I think the presence of "Fabric" in the class name causes people to believe it is some class provided by the API, rather than a class from their mod/the example mod.

dicedpixels:
fix #162

The current mod ID `fabric-docs-reference` and its Java form `FabricDocsReference` have been a confusion for new mod developers. Many have assumed it's one of Fabric's inbuilt class/constant.

This PR changes the mod ID to `example-mod` and `ExampleMod`. There were a couple of `fabric_docs_reference` which I have changed to `example-mod` as well.

I have tested many things best to my ability, as well as verify the docs as much as possible.

I'll leave this PR as a draft for comments/ larger change requests for a while, given the size of it.


its-miroma:
TODO: check references to `mod-id` and `mod_id` also.

dicedpixels:
Did some quick tests on the 1.20.4 and 1.21 reference mods as well. Didn't notice anything broken.

modmuss50 (approved):

modmuss50:
Impossible to review, I trust its correct. Feel free to merge and fix any issues in a followup pr.

cassiancc (approved):
Echoing modmuss, this is a good change that's impossible to review. If anything slips by, it can be fixed in a later PR.

PR #317, opened 2025-03-30
Add `.useBlockPrefixedTranslationKey()`

La-ryssa:
fix #265

PR #358, opened 2025-09-13
Call `.useBlockPrefixedTranslationKey()`

### LINKED ISSUE #265: "Creating Your First Block": missing call to `.useBlockPrefixedTranslationKey()`
If you follow the "Creating Your First Block" page by copy pasting all the available code, you will not be able to advance into the tutorial as the very first code snippet is missing .useBlockPrefixedTranslationKey() which is supposed to be in line 10.

vgl266:
fix #265 
supersede and close #317


modmuss50:
This seems to be causing the build to fail:
> java.lang.RuntimeException: Existing translation key found - block.fabric-docs-reference.condensed_dirt - Duplicate will be ignored.

vgl266:
@modmuss50 there are 2 ways to deal with it:
1. remove 
`translationBuilder.add(ModBlocks.CONDENSED_DIRT.asItem(), "Condensed Dirt");`
and similar entries from translationBuilder
2. Add reference into Adding Block Translations page that the reader needs to also add item string separately if they aren't going to use datagen later

Since aforementioned code doesn't appear to be referenced on datagen page, I propose the first solution

vgl266:
Don't really understand why it doesn't accept en_US.json

its-miroma:
@vgl266 You should run datagen again locally.

cassiancc (approved):

cassiancc:
Thank you for your contribution!

PR #324, opened 2025-04-24
Items: Item Models

Fellteros:
Heya,

I noticed that docs are missing page on item models, so I decided to create one. It's mostly copied from Minecraft Wiki (I hope that's okay, it's mentioned at the end) with a few modifications of my own. I'm a little unsure about the formatting of the page, so I'm open to any suggestions. Also, English isn't my native language, feel free to correct any sentences you think are a bit sketchy.

PR #364, opened 2025-09-21
Blocks: Block Models

Fellteros:
Since #324 got merged recently, I made a similar page for block models. It's basically the same, just for blocks


.vitepress/sidebars/develop.ts
```diff
@@ -1,4 +1,4 @@
-import { Fabric } from "../types";
+import {Fabric} from "../types";
```
its-miroma:
```suggestion
import { Fabric } from "../types";
```

develop/blocks/block-models.md
```diff
@@ -0,0 +1,246 @@
...
+3. `cullface`: _String_. Can be: `down`, `up`, `north`, `south`, `west`, or `east`. Specifies whether a face does not need to be rendered when there is a block touching it in the specified position.
+It also determines the side of the block to use the light level from for lighting the face, and if unset, defaults to the side.
+4. `rotation`: _Integer_. Rotates the texture by the specified number of degrees in 90 degree increments. Rotation does not affect which part of the texture is used.
+Instead, it amounts to permutation of the selected texture vertices (selected implicitly, or explicitly though `uv`).
+5. `tintidex`: _Integer_. Tints the texture on that face using a tint value. The default value, `-1`, indicates not to use the tint.
+Any other number is provided to `BlockColors` to get the tint value corresponding to that index (returns white when the block doesn't have a tint index defined).
+
+## Sources and Links {#sources-and-links}
+
+You can visit Minecraft Wiki's [Item Models page](https://minecraft.wiki/w/Model#Item_models) for a more detailed walkthrough. A lot of information here is from that page.
```
its-miroma:
```suggestion
You can visit Minecraft Wiki's [Block Models page](https://minecraft.wiki/w/Model#Block_models) for a more detailed walkthrough. A lot of information here is from that page.
```

develop/blocks/block-models.md
```diff
@@ -0,0 +1,246 @@
...
+
+### Display {#display}
+
+```json
+{
+  "display": {
+    "<position>": {
+      "rotation": [0, 0, 0],
+      "translation": [0, 0, 0],
+      "scale": [0, 0, 0]
```
its-miroma:
```suggestion
      "rotation": [0.0, 0.0, 0.0],
      "translation": [0.0, 0.0, 0.0],
      "scale": [0.0, 0.0, 0.0]
```

develop/blocks/block-models.md
```diff
@@ -0,0 +1,246 @@
...
+
+```json
+{
+  "from": [0.0, 0.0, 0.0],
+  "to": [0.0, 0.0, 0.0]
+}
+```
+
+`from` specifies the starting point of the cuboid according to the scheme `[x, y, z]`, relative to the lower left corner. `to` specifies the ending point. A cuboid as big as a standard block would start at `[0, 0, 0]` and end at `[16, 16, 16]`.
+The values of both must be between **-16** and **32**, which means that every block model can be at most 3x3 blocks big.
```
its-miroma:
```suggestion
The values of both must be between **-16** and **32**, which means that every block model can be at most 3×3 blocks big.
```

develop/blocks/block-models.md
```diff
@@ -0,0 +1,246 @@
...
+    "rescale": "true/false"
+  }
+}
+```
+
+`rotation` defines the rotation of an element. It contains four more values:
+
+1. `origin`: _Three floats_. Sets the center of the rotation according to the scheme `[x, y, z]`.
+2. `axis`: _String_. Specifies the direction of rotation, and must be one of these: `x`, `y` and `z`.
+3. `angle`: _Float_. Specifies the angle of rotation. Ranges from **-45** to **45**.
```
its-miroma, line 109: Item models says:

```suggestion
3. `angle`: _Float_. Specifies the angle of rotation. Ranges from **-45** to **45** in 22.5 degree increments.
```

I'd guess this also applies here ?

develop/blocks/block-models.md
```diff
@@ -0,0 +1,246 @@
...
+      "texture": "...",
+      "cullface": "...",
+      "rotation": 0,
+      "tintindex": 0
+    }
+  }
+}
+```
+
+`faces` holds all the faces of the cuboid. If a face is not set, it will not be rendered. Its keys can be one of: `down`, `up`, `north`, `south`, `west` or `east`. Each key contains the properties for that face:
```
its-miroma: Could you also fix a spelling mistake in Item Models at line 242 (Each key_**s**_)?

develop/blocks/block-models.md
```diff
@@ -0,0 +1,246 @@
...
+    "particle": "...",
+    "<texture_variable>": "..."
+  }
+}
+```
+
+The `textures` tag holds the textures of the model, in the form of an identifier or a texture variable. It contains three additional objects:
+
+1. `particle`: _String_. Defines the texture to load particles from. This texture is also used as an overlay if you are in a nether portal, and used for water and lava's still textures. Is also considered a texture variable that can be referenced as `#particle`
+2. `<texture_variable>`: _String_. It creates a variable and assigns a texture. Can be further referenced with the `#` prefix (e.g., `"top": "namespace:path"` ⇒ `#top`)
```
its-miroma:
```suggestion
2. `<texture_variable>`: _String_. It creates a variable and assigns a texture. Can be later referenced with the `#` prefix (e.g., `"top": "namespace:path"` ⇒ `#top`)
```

its-miroma (changes requested):
Thank you!

I don't really like that we're essentially duplicating a page twice, with very small changes. The oversight at the end of the file proves how fragile this is.

We might want to consider doing something like what's been done in [Player Guides: Updating Fabric](https://github.com/FabricMC/fabric-docs/blob/a01712fe00e47b2e1a4d8fb08e37393b3fa6491e/players/updating-fabric.md?plain=1#L19):

<https://github.com/FabricMC/fabric-docs/blob/a01712fe00e47b2e1a4d8fb08e37393b3fa6491e/players/updating-fabric.md?plain=1#L19>

develop/blocks/block-models.md
Fellteros: In 1.21.6 (i think?) this was removed

Fellteros:
Thanks for the fast replies, I'll try the approach you suggested and see how it goes.

develop/blocks/block-models.md
```diff
@@ -0,0 +1,246 @@
...
+
+```json
+{
+  "rotation": [0.0, 0.0, 0.0],
+  "translation": [0.0, 0.0, 0.0],
+  "scale": [0.0, 0.0, 0.0]
+}
+```
+
+1. `rotation`: _Three floats_. Specifies the rotation of the model according to the scheme `[x, y, z]`.
```
Fellteros: @its-miroma What do you think about adding non-breaking spaces to the "[x, y, z]" part? It looks weird when it's soft wrapped (see the image). I know that there is a lint rule that forbids this, but in this case, I think it would be reasonable

Image

its-miroma: You may do it and ignore the false positive, I'll push an exception to the markdownlint config.

But I think that flipping the sentence structure might fix this issue and be clearer, something like:

```md
1. `rotation` (`float [x, y, z]`): Specifies the rotation of the model.
```

Fellteros: Well there are other places where that can't be applied, so I think I'll stick with my original idea. thx for the suggestion though

its-miroma (approved):

develop/blocks/block-models.md
```diff
@@ -0,0 +1,124 @@
+---
+title: Block Models
+description: A guide to writing and understanding block models.
+authors:
+  - Fellteros
+  - its-miroma
+---
+
```
its-miroma, line 8:
```suggestion

<!-- markdownlint-disable search-replace -->

```

develop/items/item-models.md
```diff
@@ -3,6 +3,7 @@ title: Item Models
 description: A guide to writing and understanding item models.
 authors:
   - Fellteros
+  - its-miroma
   - VatinMc
 ---
```
its-miroma, line 9:
```suggestion

<!-- markdownlint-disable search-replace -->

```

its-miroma (changes requested):
disable linting rule

its-miroma (approved):

modmuss50 (approved):

PR #367, opened 2025-09-30
Notes on Creating a Project

cassiancc:
New users are prone to skipping ahead and attempting to create a project through IntelliJ Idea. This causes issues, as Minecraft Development's plugins are not always kept up to date (see recent posts in the Fabric Discord confused why unpick isn't working). New users should be recommended toward the existing tutorial.


kevinthegreat1 (approved):
Maybe using a warning would be better, but lgtm.

cassiancc:
It could be a warning in Creating a Project, but I feel having it here in the IntelliJ docs is better and having a floating warning would look odd.

PR #371, opened 2025-10-14
Switch to Rolldown

kevinthegreat1:
Rollup is a dependency of vite, which is a dependency of vitepress. rolldown is a beta rust rewrite of rollup and is planned to replace rollup. This pr switches vite to rolldown-vite, the beta version of vite using rolldown which is planned to replace vite. 

Using rolldown gives a significant performance improvement, and allows us to build the entire website in less than 3GB of ram (including the versions I added back). This pr also adds back versions previously deleted, namely 1.21.4 and 1.21.8.

I believe the usage of beta software is justified in this case, as it is planned to replace its respective packages.


modmuss50:
Can you split this PR to only switch to rolldown? its impossible to review like this.

kevinthegreat1:
Done. Sorry about that.

modmuss50:
Thanks, I'll take a look shortly. As you guessed I'm not super keen on using beta software, but if it's going to get improvements/fixes faster than what we have now that can only be a good thing.

modmuss50 (approved):

modmuss50:
Lets try it, worse case we roll back.

PR #374, opened 2025-10-16
Component updates

its-miroma:


kevinthegreat1 (dismissed):

kevinthegreat1:
Still breaks when I reload. I don’t know why this is happening cause I don’t know vue.

its-miroma:
Currently I decided to give up and wait for tomorrow morning to have a fresh mind

coz I cannot repro locally so I have to push every time and wait for it to build and it's annoying

.vitepress/theme/components/AuthorsComponent.vue
```diff
@@ -3,98 +3,87 @@ import { useData } from "vitepress";
...
+          author.noGitHub
+            ? '/assets/avatater.png'
+            : 'https://wsrv.nl/?af&maxage=7d&url=' +
+              encodeURIComponent(`https://github.com/${author.name}.png?size=32`)
+        "
+        :alt="author.name"
+        loading="lazy"
+        @error="onError"
+      />
+    </component>
```
kevinthegreat1, line 51: From a quick look, why don't you just make it use `avatater.png` if `deletedGithub` or `noGithub`? What am I missing?

its-miroma: We do not keep a list of users who have deleted their GitHub; this @error logic triggers if the image fails to load.

its-miroma: We do not keep a list of users who have deleted their GitHub; this @error logic triggers if the image fails to load.

.vitepress/config.ts
```diff
@@ -90,7 +90,7 @@ export default defineVersionedConfig(
       transformItems,
     },
 
-    srcExclude: ["README.md"],
+    srcExclude: ["README.md", ...(typeof env === "number" ? ["versions"] : [])],
```
IMB11, line 93: I think PR builds which touch versioned content or the versioning system should probably have versioning enabled no?

its-miroma: I considered that: either I must call git diff with upstream (which means adding the remote in ci and fetching it), or set an environment variable depending on modified files (which I can do in GitHub Actions but idk about Netlify), or fetch the labels on the GitHub PR for if they include `new-version` (but labels are not set automatically so the first push will not include versions).
I concluded that new versions are not common enough to warrant adding all this overhead; besides with new versions, since update.ts is stable, no changes or surprises are expected and no one really reviews the old version. 
Correct me if I'm wrong, or if I've forgotten any other option.

IMB11 (dismissed):
Looks reasonable to me - lgtm if others are happy with it.

IMB11:
Ah, one thing I have noticed, if we're disabling versioning for PRs we should hide the version switcher component.

modmuss50 (approved):

PR #380, opened 2025-10-29
revise dependencies

its-miroma:
- use `tinyglobby` instead of `glob`
  - this is what `vitepress` uses internally
  - it's now a dependency because I'll use it outside of update.ts
- remove unnecessary `math` dependency
- override `vitepress`
  - `vvp` declares an old version, with conflicting types


modmuss50 (commented):
Removing depedencies can only be a good thing 👍

modmuss50 (approved):

PR #382, opened 2025-10-29
format

its-miroma:
this pr exists to limit the size of #374
#388 should be merged to also get the changes to ChoiceComponent

- prettier:
```yaml
experimentalOperatorPosition: start
printWidth: 100
quoteProps: consistent
trailingComma: es5
```
- use arrow functions instead of `function`
- shorten some method and variable names
- prefer template strings to string concatenation
- use top-level await in update.ts
- report dead links outside of translated pages

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+max_line_length = 100
+trim_trailing_whitespace = true
+
+[*.java]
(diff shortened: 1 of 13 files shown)


IMB11:
Good, but I would move this into an editor config file rather than prettier, then we can configure other languages if needed, for now just scope it to JS/TS and Vue stuff

IMB11 (dismissed):
Looks fine to me, lgtm

kevinthegreat1 (approved):
LGTM, seems like reasonable styling choices based on what we have.

cassiancc (approved):

PR #408, opened 2025-11-19
Warnings for Yarn Mappings

cassiancc:
Versions prior to 1.21.10 are written with Yarn Mappings, which has caused some confusion for users who have made new projects on old versions that now default to Mojmap. For clarity, expanding the warning to also mention the mapping set should be added.

I've opted to not warn for Mojang Mappings as discussed on Discord, especially as this would only apply to 1.21.10/11 with future versions being unobf.


.vitepress/types.ts
```diff
@@ -126,10 +126,15 @@ export namespace Fabric {
        */
       latestVersion: string;
 
-      /**
+        /**
```
its-miroma:
```suggestion
      /**
```

its-miroma (changes requested):
LGTM

its-miroma (approved):

PR #417, opened 2025-11-29
Fix Incorrect Section About Event Callbacks

### LINKED ISSUE #416: "develop/events" contains inaccuracies relating to event packages and interface names
As brought up by d_iscord_made_me_do_it on Discord after a user became confused, the Events page currently displays the following:

Image

However, it is not a hard rule that event interfaces be named <EventName>Callback, and it is outright false that Fabric API events are centralized in a single package.

As it was pointed out, it may be possible that someone used an LLM to generate the inaccurate package information, so it may also be relevant looking into who contributed that portion's commits and check if there are other inaccuracies or LLM-esque issues.

Will likely PR a fix myself in the foreseeable future unless somebody beats me to the punch.

dicedpixels:
As discussed here: [#docs](https://discord.com/channels/507304429255393322/1208846408552030238/1444143328785338518)

Closes https://github.com/FabricMC/fabric-docs/issues/416

Fabric API itself seems to be inconsistent when it comes to callback naming. So the line about naming convention is removed. 

Since not all events are in a single `event` package, that line is also removed.

### DIFF
--- a/develop/events.md
+++ b/develop/events.md
@@ -31,9 +31,7 @@ Callbacks are a piece of code that is passed as an argument to an event. When th
 
 ### Callback Interfaces {#callback-interfaces}
 
-Each event has a corresponding callback interface, conventionally named `<EventName>Callback`. Callbacks are registered by calling `register()` method on an event instance, with an instance of the callback interface as the argument.
-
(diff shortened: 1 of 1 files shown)


ekulxam (approved):

ekulxam:
Do we need to add a line explaining how to find all existing events?

its-miroma (approved):

PR #419, opened 2025-12-02
Add page for Java bytecode

Earthcomputer:


develop/mixins/bytecode.md
```diff
@@ -0,0 +1,68 @@
...
+description: Learn the basics of Java bytecode so that you can write mixins more effectively.
+authors:
+- Earthcomputer
+---
+
+Mixins operate on Java bytecode, so having a basic grasp on the fundamentals of how Java bytecode works is essential to understanding Mixins.
+
+## Viewing the Bytecode of a Class {#show-bytecode}
+
+You can view the bytecode of a library class (such as a Minecraft class) in IntelliJ by opening the class in the editor, then selecting `Show Bytecode` inside the `View` menu. This shows you exactly what is going on there in a way the decompiled code will not.
```
cassiancc: This would benefit from images.

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,68 @@
+---
+title: Java Bytecode
+description: Learn the basics of Java bytecode so that you can write mixins more effectively.
+authors:
+- Earthcomputer
+---
+
+Mixins operate on Java bytecode, so having a basic grasp on the fundamentals of how Java bytecode works is essential to understanding Mixins.
```
cassiancc: This article goes on to describe the process of viewing bytecode in IntelliJ Idea specifically. I would recommend either merging or linking to the [Visual Studio Code](https://deploy-preview-419--nimble-elf-d9d491.netlify.app/develop/getting-started/vscode/tips-and-tricks#viewing-bytecode) explanation on viewing bytecode as well, since the docs should currently cover both.

cassiancc (changes requested):

develop/mixins/bytecode.md
Earthcomputer: There is no way to do this with vscode. In general I would extremely not recommend using vscode for mixins anyway.

dicedpixels: Cassian meant doing this in a non-IDE specific way. The VSvCode guide has a `javap` example.

develop/getting-started/intellij-idea/tips-and-tricks.md
```diff
@@ -293,6 +294,14 @@ Minecraft's `ScreenHandler` class has some examples. To toggle the render view,
 
 ![JavaDoc editing](/assets/develop/misc/using-the-ide/comments_07.png)
 
+## Viewing Bytecode {#viewing-bytecode}
+
+Viewing bytecode is necessary when writing mixins. You can view the bytecode of a library class (such as a Minecraft class) in IntelliJ by opening the class in the editor, then selecting `Show Bytecode` inside the `View` menu.
```
its-miroma:
```suggestion
Viewing bytecode is necessary when writing mixins. You can view the bytecode of a library class (such as a Minecraft class) by opening the class in the editor, then selecting `Show Bytecode` inside the `View` menu.
```

develop/getting-started/intellij-idea/tips-and-tricks.md
```diff
@@ -293,6 +294,14 @@ Minecraft's `ScreenHandler` class has some examples. To toggle the render view,
...
+
+![Show bytecode 1](/assets/develop/getting-started/intellij/show_bytecode_1.png)
```
its-miroma:
```suggestion
!["Show Bytecode" button in the "View" menu](/assets/develop/getting-started/intellij/show_bytecode.png)
```

develop/getting-started/intellij-idea/tips-and-tricks.md
```diff
@@ -293,6 +294,14 @@ Minecraft's `ScreenHandler` class has some examples. To toggle the render view,
...
+
+![Show bytecode 2](/assets/develop/getting-started/intellij/show_bytecode_2.png)
```
its-miroma:
```suggestion
![Bytecode of `java/lang/String.class`](/assets/develop/getting-started/intellij/bytecode_example.png)
```

public/assets/develop/getting-started/intellij/show_bytecode.png
its-miroma, line 1:
```sh
mv show_bytecode_1.png show_bytecode.png
```

public/assets/develop/getting-started/intellij/bytecode_example.png
its-miroma, line 1:
```sh
mv show_bytecode_2.png bytecode_example.png
```

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,113 @@
+---
+title: Java Bytecode
+description: Learn the basics of Java bytecode so that you can write mixins more effectively.
```
its-miroma:
```suggestion
description: Learn about Java bytecode, so that you can write mixins effectively.
```

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,113 @@
...
+authors:
+- Earthcomputer
+---
+
+Mixins operate on Java bytecode, so having a basic grasp on the fundamentals of how Java bytecode works is essential to understanding Mixins.
```
its-miroma:
```suggestion
Mixins operate on Java bytecode, so to understand them one needs a grasp on the fundamentals of Java bytecode.
```

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Viewing the Bytecode of a Class {#viewing-bytecode}
+
+If can look at the bytecode of any Minecraft/library class using your IDE. This shows you exactly what is going on there in a way the decompiled code will not.
```
its-miroma:
```suggestion
You can look at the bytecode of any Minecraft/library class using your IDE. This shows you exactly what is going on there, whereas the decompiled code might not.
```

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,113 @@
...
+
+To find out how view the bytecode of a class, please consult the [tips and tricks](../getting-started/tips-and-tricks) article for your IDE.
```
its-miroma:
```suggestion
To find out how to see the bytecode of a class, please consult the Viewing Bytecode section of the [Tips and Tricks page](../getting-started/tips-and-tricks) for your IDE.
```

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Names and Symbols {#names-and-symbols}
+
+In bytecode, many things like classes, fields, and methods, are still identified by the name (and descriptor for fields and methods, we'll get to that), just like they are in Java. However the exact format of those names differ a little bit.
```
its-miroma:
```suggestion
In bytecode, many things like classes, fields, and methods, are still identified by the name (and the descriptor for fields and methods, we'll get to that), just like they are in source code. However the exact format of those names differs a little bit.
```

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Class Names {#class-names}
+
+Classes are generally referred to by their _internal name_, which is roughly equivalent to the fully qualified class name (full name including the package), except that all `.` dots are replaced with `/` slashes. For example, the internal name of the `java.lang.Object` class is `java/lang/Object`.
```
its-miroma:
```suggestion
Classes are generally referred to by their _internal name_, which is roughly equivalent to the fully qualified class name (full name including the package), where all dots `.` are replaced with slashes `/`. For example, the internal name of the `java.lang.Object` class is `java/lang/Object`.
```

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,378 @@
...
+
+Nested classes use a `$` symbol to separate their name from their enclosing class. For example, if you have a class named `Inner` inside a class named `Foo` inside a package named `pkg`, then the internal name of `Inner` would be `pkg/Foo$Inner`.
```
its-miroma:
````suggestion
Nested classes use `$` symbols to separate their name from the enclosing classes. For example, given:

```java
package pkg;

class Foo {
    class Bar {
    }
}
```

... the internal name of `Bar` would be `pkg/Foo$Bar`.
````

correct me if I'm wrong, but would triple nesting make it `pkg/Foo$Bar$Baz`? no need to specify it, just to see if my wording is accurate.

also, we use tabs for indenting Java, but I cannot add tab characters to the code block through GitHub. you may want to change that in your IDE, but I'm not sure if the linter would accept them.

File: develop/mixins/bytecode.md [hunk 13]
```diff
@@ -0,0 +1,378 @@
...
+
+Anonymous classes are similar to nested classes, except that their names are simply a number. The first anonymous class declared within a class has the name `1`, the second has the name `2`, and so on. So if there was an anonymous class inside the `Foo` class above, its name might be `pkg/Foo$1`.
```
its-miroma:
```suggestion
Anonymous classes are internally named like nested classes, but instead of a class name they use numbers. For example, if there were two anonymous classes in the `Foo` class from the code block above, their internal names would be `pkg/Foo$1` and `pkg/Foo$2` respectively.
```

develop/mixins/bytecode.md
```diff
@@ -7,15 +7,11 @@ authors:
 
 Mixins operate on Java bytecode, so having a basic grasp on the fundamentals of how Java bytecode works is essential to understanding Mixins.
 
-## Viewing the Bytecode of a Class {#show-bytecode}
+## Viewing the Bytecode of a Class {#viewing-bytecode}
 
-You can view the bytecode of a library class (such as a Minecraft class) in IntelliJ by opening the class in the editor, then selecting `Show Bytecode` inside the `View` menu. This shows you exactly what is going on there in a way the decompiled code will not.
+If can look at the bytecode of any Minecraft/library class using your IDE. This shows you exactly what is going on there in a way the decompiled code will not.
```
MildestToucan: "If can look at" doesn't sound right.

MildestToucan (changes requested):
Likely typo should be fixed before merging

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,378 @@
...
+
+Nested classes use a `$` symbol to separate their name from their enclosing class. For example, if you have a class named `Inner` inside a class named `Foo` inside a package named `pkg`, then the internal name of `Inner` would be `pkg/Foo$Inner`.
+
+Anonymous classes are similar to nested classes, except that their names are simply a number. The first anonymous class declared within a class has the name `1`, the second has the name `2`, and so on. So if there was an anonymous class inside the `Foo` class above, its name might be `pkg/Foo$1`.
+
+Local classes (classes defined within a method), have a number followed by the name. For example, a local class name might look like `pkg/Foo$1Local`.
+
+### Type Descriptors {#type-descriptors}
+
+Sometimes bytecode needs to refer to a type which could be a primitive type or array type, not only a class type. In this situations, _type descriptors_ are used. Here is a table of data types and how they map to their type descriptors:
```
its-miroma:
```suggestion
When bytecode needs to refer to primitive types or arrays, _type descriptors_ are used. Here is a table of data types and their respective type descriptors:
```

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,378 @@
...
+| `boolean` | `Z`                                                                  |
+| `byte`    | `B`                                                                  |
+| `char`    | `C`                                                                  |
+| `double`  | `D`                                                                  |
+| `float`   | `F`                                                                  |
+| `int`     | `I`                                                                  |
+| `long`    | `J`                                                                  |
+| `short`   | `S`                                                                  |
+| `void`    | `V`                                                                  |
+| Arrays    | `[` + the element type, e.g. `int[]` -> `[I`                         |
```
its-miroma:
```suggestion
| Arrays    | `[` + the element type: `int[]` -> `[I`                         |
```

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,378 @@
...
+| Objects   | `L` + the internal name + `;`, e.g. `String` -> `Ljava/lang/String;` |
```
its-miroma:
```suggestion
| Objects   | `L` + the internal name + `;`: `String` -> `Ljava/lang/String;` |
```

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,378 @@
...
+
+### Field and Method Descriptors {#field-and-method-descriptors}
+
+In bytecode, fields and methods are identified by the comibination of their name and _descriptor_. A field descriptor is simply the type descriptor of the data type of the field.
```
its-miroma:
```suggestion
In bytecode, fields and methods are identified by combining their name and _type descriptor_.
```

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,378 @@
...
+
+For methods, the descriptor is the combination of the parameter types and the return type of the method. Here is an example of a method descriptor for the following method:
```
its-miroma:
```suggestion
Descriptors for methods combine the parameter types and the return type. For example, the following method:
```

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,378 @@
...
+void drawText(int x, int y, String text, int color) {
+    // ...
+}
+```
+
+```text
+(IILjava/lang/String;I)V
+```
+
+You may notice that it consists of the parameter types inside parentheses, followed by the return type, in this case `V` for `void`. The descriptors for the parameter types are directly concatenated together with no separator. In this case, there are two copies of `I` for `int`, then `Ljava/lang/String;` for the `String`, and then one more `I` for the last `int`.
```
its-miroma:
```suggestion
... has descriptor `(IILjava/lang/String;I)V`.

The descriptors for the parameter types are directly concatenated together, with no separator. In this case, there is `I` for `int` twice (both `x` and `y`), then `Ljava/lang/String;` for `String` (`text`), and one more `I` for the last `int` (`color`).
```

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,378 @@
...
+public Vec3 add(double x, double y, double z) {
+    // ...
+}
+```
+
+The `x` parameter has index 1, the `y` parameter has index 3, and the `z` parameter has index 5.
+
+In static methods, there is no `this` object, so other variables start counting at 0 instead of 1.
+
+Even though local variables are identified by index, many libraries retain debug information containing the names of the local variables as well. This includes Minecraft since the unobfuscated releases started in version 26.1 (upcoming). Mixin is able to use this debug information to allow you to target local variables by name.
```
cassiancc: As the docs are versioned, information should primarily focus on the version in use.

```suggestion
Even though local variables are identified by index, many libraries retain debug information containing the names of the local variables as well. This does not yet include Minecraft, as it will remain obfuscated until the upcoming version 26.1. Mixin is able to use this debug information to allow you to target local variables by name.
```

cassiancc (changes requested):

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,378 @@
...
+
+```text
+(IILjava/lang/String;I)V
+```
+
+You may notice that it consists of the parameter types inside parentheses, followed by the return type, in this case `V` for `void`. The descriptors for the parameter types are directly concatenated together with no separator. In this case, there are two copies of `I` for `int`, then `Ljava/lang/String;` for the `String`, and then one more `I` for the last `int`.
+
+### Constructors and Static Initializers {#constructors-and-static-initializers}
+
+At the bytecode level, constructors are just another method, and are special only in ways beyond the scope of this overview. A constructor's method name is `<init>` (with the `<>` angled brackets), and the return type in its descriptor is `V` (`void`). All non-static field initializers (the part after the `=` sign in a field declaration) are compiled into the `<init>` methods of a class and assigned to the field there.
```
its-miroma:
```suggestion
At the bytecode level, constructors are just another method: further differences between the two fall beyond the scope of this overview.

A constructor's method name is `<init>` (with the `<>` angled brackets), and the return type in its descriptor is `V` (`void`). All non-`static` field initializations, after compilation, will be found inside of the `<init>` methods.
```

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,378 @@
...
+
+Static initializers are the `static {}` block in Java that is run when a class is loaded. It's also just another method, its name is `<clinit>` and its descriptor is `()V`. Just as for non-static fields and constructors, static field initializers (with some exceptions) are compiled into the `<clinit>` method of a class.
```
its-miroma:
```suggestion
On the other hand, `static` initializers form the `static {}` block in Java, which is run when a class is loaded. It's also just another method: its name is `<clinit>`, and its descriptor is `()V`. Just like non-`static` fields and constructors, `static` field initializers (with some exceptions) are compiled into the `<clinit>` method of a class.
```

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,378 @@
...
+
+Longs and doubles take up 2 indexes in the LVT. This means that in the following method:
+
+```java
+public Vec3 add(double x, double y, double z) {
+    // ...
+}
+```
+
+The `x` parameter has index 1, the `y` parameter has index 3, and the `z` parameter has index 5.
```
its-miroma:
```suggestion
... the `x` parameter gets index 1, the `y` parameter gets index 3, and the `z` parameter gets index 5.
```

its-miroma (changes requested):
I'm not done reviewing the whole article, but I'll send what I've done till now. 

I am also researching which code block language can most closely reflect Java bytecode (instead of text), and I'll try adding mermaid.

develop/mixins/bytecode.md
Earthcomputer: Yes, it uses `$` for each level of nesting.

develop/mixins/bytecode.md
Earthcomputer: "behave like" -> "are named similarly"

it's the naming we are referring to, not the behaviour

develop/mixins/bytecode.md
Earthcomputer: No need to use the fancy arrow character, seems vitepress already turns `->` into that

develop/mixins/bytecode.md
its-miroma: Of course 👍

develop/mixins/bytecode.md
Earthcomputer: > In bytecode, fields and methods are identified by combining their name and _type descriptor_.

Method descriptors are not a consequence of field descriptors

develop/mixins/bytecode.md
Earthcomputer: "The differences between the two" -> "the detailed differences between the two"

develop/mixins/bytecode.md
its-miroma: great! i did not know that

develop/mixins/bytecode.md
Earthcomputer: ->

> On the other hand, static initializers (the `static {}` block in Java)

The word "form" felt weird to me here, I just wanted to clarify what a static initializer was, so I switched to clarifying it in brackets. I also undid the monospaced font for static in this one instance, as I wanted to refer to static initializers as a concept rather than a specific code construct.

develop/mixins/bytecode.md
its-miroma: not sure what i was thinking 😅 i guess i read parameters

its-miroma: not sure what i was thinking 😅 i guess i read parameters

its-miroma: not sure what i was thinking 😅 i guess i read parameters

its-miroma: not sure what i was thinking 😅 i guess i read parameters

its-miroma: not sure what i was thinking 😅 i guess i read parameters

develop/mixins/bytecode.md
its-miroma: @Earthcomputer you applied this suggestion in the wrong place

Earthcomputer: Fixed

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,392 @@
...
+
+Pops the top two values off the operand stack, adds them, and pushes the result onto the operand stack. After this instruction is finished, the operand stack is as follows:
+
+| Operand Stack |
+|---------------|
+| 47            |
+
+**istore 2**
+
+Pops the top value of the operand stack and assigns it to the variable in LVT slot 2. After this instruction is finished, the operand stack is empty and the variable in LVT slot 2 has a value of 47.
```
ekulxam: Would it be helpful to show the LVT slots alongside the operand stack?

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,392 @@
...
+- `iadd`, `ladd`, `fadd`, `dadd`: addition.
+- `isub`, `lsub`, `fsub`, `dsub`: subtraction.
+- `imul`, `lmul`, `fmul`, `dmul`: multiplication.
+- `idiv`, `ldiv`, `fdiv`, `ddiv`: division.
+- `irem`, `lrem`, `frem`, `drem`: modulo.
+- `ineg`, `lneg`, `fneg`, `dneg`: negation. Pops one value off the stack rather than two.
+
+### Returns {#return-instructions}
+
+Return out of the method, with the value on top of the operand stack (except in the case of `void`). Prefixed with `i`, `l`, `f`, `d`, and `a` in the same way as [variable instructions](#variable-instructions), except for `void` for which the instruction is simply `return`.
```
ekulxam: > Prefixed with `i`, `l`, `f`, `d`, and `a`

Do we need to put this earlier? I’m asking because the previous section also has these (`iadd`).

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,392 @@
...
+
+static lambda$hello$1 ()V
+  getstatic System.out
+  ldc "Hello, World!"
+  invokevirtual println
+
+  return
+```
+
+You can see here that the contents of the lambda has been moved into a separate method, in this case `lambda$hello$1`. If you want to target the contents of a lambda with Mixin, this is the method you'll want to be targetting. The lambda instance is then created with the `invokedynamic` instruction and then stored into the variable `r`.
```
ekulxam: I believe targetting is actually spelled targeting.

ekulxam (changes requested):

develop/mixins/bytecode.md
Earthcomputer: Wiktionary lists it as a rare alternative spelling. Some sources list it as only valid in British English so I'll change it. British English does this double consonant thing in the present participle more often than American English.

its-miroma: In English (with some exceptions and differences between American and British irrelevant to this case), one condition for doubling the final consonant is it must be in the stressed syllable[^1]:

**get** becomes getting because the t is in the stressed syllable (same for **put**(ting), for**get**(ting), un**set**(ting), reg**ret**(ting)...).

the second t in "**tar**get" is not stressed, so it's not duplicated (same for **mar**ket(ing), **par**rot(ing), **bud**get(ing)...).

[^1]: https://english.stackexchange.com/questions/205815/is-targetted-a-standard-british-english-spelling

> Wiktionary[^2] lists it as a rare alternative spelling

[^2]: https://en.wiktionary.org/wiki/targetting

because language rules get broken all the time, and grammarians have no authority to enforce them. that's why no one can tell you "who are you taking to?" is inaccurate because **whom** is correct: that's no longer followed or relevant in non-formal speech or writing[^3].

[^3]: https://motivatedgrammar.wordpress.com/2012/09/10/back-to-school-review-who-and-whom/

targeting is correct, and most contemporary style guide will say targetting is a misspelling, but who knows what the future holds for us.

I don't personally like Wiktionary's optimistic definition of it as an alternative ("something which can be chosen instead"[^4]), because I don't think it's mainstream enough to be that. Even the oxymoron of rare + alternative hints that their wording is weak imo.

[^4]: https://www.merriam-webster.com/dictionary/alternative

it's late where I'm at, good night everyone!

develop/mixins/bytecode.md
Earthcomputer: Is it difficult to follow as it is? This section is about the operand stack, not the LVT

develop/mixins/bytecode.md
Earthcomputer: I added something in the previous section about it

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,394 @@
...
+    int result = this.x + offset;
+    return result;
+}
+```
+
+This would compile to something like the following:
+
+```text
+public getX (I)I
+  iload 0  # this
```
Tenneb22:
```suggestion
  aload 0  # this
```
I think this needs to be `aload`, like in the example below

Earthcomputer: Thanks, I noticed this mistake but only fixed it in one of the two places

Earthcomputer:
One more thing to add - constructor calls

cassiancc (dismissed):
My issues have all been addressed, I'll let Miroma take it from here.

dicedpixels (dismissed):

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,504 @@
...
+
+### Returns {#return-instructions}
+
+Return instructions close the method call, returning the value at the top of the Operand Stack (except for `void` methods).
+
+When prefixed with `i`, `l`, `f`, `d`, and `a`, just like [variable instructions](#variable-instructions), the method returns a value of that type. The instruction for `void` methods is simply `return`.
+
+### New Object Creation {#new-object-creation}
+
+In source code, you can write `new MyClass()`, which creates a new instance of `MyClass` and calls the constructor. In bytecode these are two distinct operations. Take the following code, for example:
```
its-miroma:
```suggestion
In source code, writing `new MyClass()` creates a new instance of `MyClass` and calls its constructor. In bytecode, these two steps become distinct operations. Take the following code, for example:
```

develop/mixins/bytecode.md
```diff
@@ -0,0 +1,504 @@
...
+  new net/minecraft/world/entity/monster/Creeper
+  dup
+  aload 0  # level
+  invokespecial net/minecraft/world/entity/monster/Creeper.<init> (Lnet/minecraft/world/level/Level;)V
+  areturn
+```
+
+:::
+
+Here, the `new` instruction creates an uninitialized instance of `Creeper`, and pushes a pointer to it onto the Operand Stack. The `dup` instruction duplicates the top of the Operand Stack so that there are two copies of the pointer to the new `Creeper` instance. Next, the arguments to the constructor are pushed onto the Operand Stack (in this case, the local variable called `level`). The `invokespecial` instruction is then used to call the constructor of the `Creeper` object. Like a regular non-static method call, this pops all the arguments and the instance the method is called on off the Operand Stack, and because the method returns `void`, nothing is pushed onto the Operand Stack afterwards. This consumes one of the pointers to `Creeper`, which is why we had to `dup` it before, as we now have one pointer remaining. Finally, we return the pointer to the `Creeper` that remains on the operand stack with the `areturn` instruction.
```
its-miroma:
````suggestion
Let’s examine what happens on the Operand Stack.

::: tabs

== Start

| Index | Local Variable Table | Operand Stack |
|-------|----------------------|---------------|
| 2     |                      |               |
| 1     |                      |               |
| 0     | `level`              |               |

LVT slot 0 contains `level`. It does not contain `this` because the method is static.

== new Creeper

| Index | Local Variable Table | Operand Stack    |
|-------|----------------------|------------------|
| 2     |                      |                  |
| 1     |                      |                  |
| 0     | `level`              | uninit `Creeper` |

Allocates an uninitialized object, and pushes a reference to it onto the Operand Stack.

== dup

| Index | Local Variable Table | Operand Stack    |
|-------|----------------------|------------------|
| 2     |                      |                  |
| 1     |                      | uninit `Creeper` |
| 0     | `level`              | uninit `Creeper` |

Duplicates the top value of the Operand Stack.

We need the Operand Stack to contain two pointers to the same object because one will be consumed by `invokespecial <init>`, and the other will be returned by `areturn`.

== aload 0

| Index | Local Variable Table | Operand Stack    |
|-------|----------------------|------------------|
| 2     |                      | `level`          |
| 1     |                      | uninit `Creeper` |
| 0     | `level`              | uninit `Creeper` |

Loads the variable from LVT slot 0 (`level`), and pushes its value onto the Operand Stack.

== invokespecial <init>

| Index | Local Variable Table | Operand Stack |
|-------|----------------------|---------------|
| 2     |                      |               |
| 1     |                      |               |
| 0     | `level`              | `Creeper`     |

Pops the parameters that will be fed into `Creeper.<init>` (in this case `level`), and the pointer to the object that will be initialized, off the Operand Stack, then calls the constructor.

Because [constructors return `void`](#constructors-and-static-initializers), the Operand Stack loses one of the two pointers to `Creeper`.

== areturn

Gets the top value of the Operand Stack, and returns it.

The method returns the pointer to the `Creeper` instance.

:::
````

- I'm not sure if the wording "uninit `Creeper`" is correct. Maybe `Object`?
- Should #constructors-and-static-initializers and #new-object-creation be merged? are they different enough?

**Review by its-miroma — CHANGES_REQUESTED** [review]

File: develop/mixins/bytecode.md (hunk 30 above)
Comment (Earthcomputer) [reply develop/mixins/bytecode.md]: I think they are different enough, one is describing how a constructor looks, the other is describing the bytecode pattern for creating a new instance of something. My original intention here was to describe the `new` instruction specifically under "common instructions", but really you should learn it as part of this bytecode pattern. Maybe it could be moved under a new section of "common bytecode patterns" along with lambdas?

Comment (its-miroma) [reply develop/mixins/bytecode.md]: I like this, but I think there's no need to distinguish between "common patterns" and "common instructions". I'd have this table of contents:

- Names and Symbols
  - ...
- Local Variables and the LVT
- Operand Stack
- Common Patterns
  - Constants
  - Variables
  - Fields
  - Operators
  - Conditionals
  - Method Invocations
  - Constructors
  - Lambdas
  - Returns

Comment (Earthcomputer) [reply develop/mixins/bytecode.md]: I'll do that but I will put lambdas and new objects at the end

**Review by its-miroma — DISMISSED** [review]

**Review by cassiancc — DISMISSED** [review]

File: develop/mixins/bytecode.md [hunk 31]
```diff
@@ -0,0 +1,567 @@
...
+- `getfield`: reads a non-static field
+- `putfield`: writes to a non-static field
+- `getstatic`: reads a static field
+- `putstatic`: writes to a static field
+
+### Method Invocations {#method-instructions}
+
+- `invokestatic`: invokes a static method
+- `invokevirtual`: invokes a non-static method. Takes polymorphism and inheritance into account, calling the overridden version where applicable
+- `invokespecial`: invokes a non-static method, exactly the one declared, without taking into account polymorphism/inheritance. Used to call constructors and superclass methods
```
MattiDragon: `invokespecial` is also used for calling private methods. Might be worth mentioning in case someone comes across it

MattiDragon (changes requested):

develop/mixins/bytecode.md
Earthcomputer: No it's not. Not anymore.

Earthcomputer: Although this could be adjusted to make it not look like an exhaustive list of examples

cassiancc (approved):

MattiDragon (approved):

Fellteros (approved):

dicedpixels (approved):

ekulxam (approved):

Tenneb22 (approved):

its-miroma (approved):

PR #424, opened 2025-12-05
Call `setDirty()` Only When Data is Changed

dicedpixels:
[As pointed out on Discord](https://discord.com/channels/507304429255393322/507982478276034570/1446407771703742576), we mark the saved data state as dirty on every load. This should be done only when the data is actually changed.

### DIFF
--- a/reference/latest/src/main/java/com/example/docs/saveddata/SavedBlockData.java
+++ b/reference/latest/src/main/java/com/example/docs/saveddata/SavedBlockData.java
@@ -43,9 +43,15 @@ public class SavedBlockData extends SavedData {
 		return blocksBroken;
 	}
 
+	// :::set_dirty
 	public void incrementBlocksBroken() {
 		blocksBroken++;
+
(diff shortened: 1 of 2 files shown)


reference/latest/src/main/java/com/example/docs/saveddata/SavedBlockData.java
```diff
@@ -43,9 +43,15 @@ public int getBlocksBroken() {
 		return blocksBroken;
 	}
 
+	// :::set_dirty
 	public void incrementBlocksBroken() {
 		blocksBroken++;
+
+		// If saved data is not marked dirty, nothing will be saved when Minecraft closes.
+		// You should only mark saved data as dirty when there was actually a change.
```
its-miroma:
```suggestion
```

already noted in markdown

its-miroma (changes requested):

its-miroma (approved):

PR #429, opened 2025-12-15
Fix Transclusions in Basic Rendering Concepts

dicedpixels:
A user pointed out that the transclusions in the Basic Rendering Concepts is broken.

I've attempted to fix it by:
* Transcluding the registration line separately.
* Transcluding the method definition separately.

Other than a couple of empty lines moving, there should be no effect to other transcludes in the page.

### DIFF
--- a/develop/rendering/basic-concepts.md
+++ b/develop/rendering/basic-concepts.md
@@ -125,7 +125,13 @@ Additionally, the code below does not fully match the explanation above: you do 
 Read the important update above for more information.
 :::
 
-@[code lang=java transcludeWith=:::1](@/reference/latest/src/client/java/com/example/docs/rendering/RenderingConceptsEntrypoint.java)
+**Element registration:**
+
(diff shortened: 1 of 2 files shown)


develop/rendering/basic-concepts.md
```diff
@@ -190,3 +196,5 @@ Let's say we want to rotate our square around the z-axis. We can do this by usin
 The result of this is the following:
 
 ![A video showing the diamond rotating around the z-axis](/assets/develop/rendering/concepts-quaternions.webp)
+
+asssssss
```
Toffikk: seems like an unintentional change? lol

dicedpixels: 😭 I use this to trigger a page refresh when I change the Java code.

cassiancc: I assumed that just didn't work and have been force closing the server this whole time, TIL.

cassiancc (approved):

PR #437, opened 2025-12-24
Custom Key Mappings

dicedpixels:
<kbd>A simple page for custom key mappings.<kbd>


develop/key-mappings.md
```diff
@@ -0,0 +1,61 @@
+---
+title: Custom Key Mappings
```
its-miroma: speaking of mappings, isn't the mojmap naming "Key Binding"?

develop/key-mappings.md
```diff
@@ -0,0 +1,61 @@
...
+description: Creating custom key mappings and reacting to them.
+authors:
+  - dicedpixels
+  - cassiancc
```
its-miroma:
```suggestion
  - cassiancc
  - dicedpixels
```

develop/key-mappings.md
```diff
@@ -0,0 +1,61 @@
...
+
+@[code lang=java transcludeWith=:::key_mapping](@/reference/latest/src/client/java/com/example/docs/keymapping/ExampleModKeyMappingsClient.java)
+
+Sticky keys can also be created with `KeyBindingHelper` by passing a `ToggleKeyMapping` instance.
+
+## Translations {#translations}
+
+You'll need to provide translations for both the key mapping and the category.
+
+![Untranslated Key Category and Mapping](/assets/develop/key-mappings/untranslated.png)
```
its-miroma, line 46: This image should be shown above the #translations heading

reference/latest/src/client/java/com/example/docs/keymapping/ExampleModKeyMappingsClient.java
```diff
@@ -0,0 +1,45 @@
...
+			ResourceLocation.fromNamespaceAndPath(ExampleMod.MOD_ID, "custom_category")
+	);
+	// :::category
+
+	// :::key_mapping
+	KeyMapping sendToChatKey = KeyBindingHelper.registerKeyBinding(
+		new KeyMapping(
+			"key.example-mod.send_to_chat", // The translation key for the key mapping.
+			InputConstants.Type.KEYSYM, // // The type of the keybinding; KEYSYM for keyboard, MOUSE for mouse.
+			GLFW.GLFW_KEY_J, // The GLFW keycode of the key.
```
its-miroma, line 28: i found these javadocs that might be a useful reference to link in the article. if you know of anything else better, do link that

https://javadoc.lwjgl.org/org/lwjgl/glfw/GLFW.html#:~:text=GLFW_KEY_0

its-miroma (changes requested):
we're never escaping the Misc category, aren't we...

reference/latest/src/client/java/com/example/docs/keymapping/ExampleModKeyMappingsClient.java
dicedpixels: I thought of linking the original source; GLFW:
https://www.glfw.org/docs/latest/group__buttons.html
https://www.glfw.org/docs/3.3/group__keys.html

However, none of these provide any additional info though. All are just constant declarations.

develop/key-mappings.md
dicedpixels: That's Yarn.

image

its-miroma: could you re-check the markdown then; there seem to be some mentions of KeyBinding, but those might be from Fabric API?

reference/latest/src/client/java/com/example/docs/keymapping/ExampleModKeyMappingsClient.java
its-miroma: I'm interested in knowing how this library behaves with keyboard layouts: if a person with an AZERTY keyboard presses A, does that register as `GLFW_KEY_Q`?

dicedpixels: > The key functions deal with physical keys, with layout independent [key tokens](https://www.glfw.org/docs/latest/group__keys.html) named after their values in the standard US keyboard layout.

https://www.glfw.org/docs/latest/group__input.html#ga1caf18159767e761185e49a3be019f8d

develop/key-mappings.md
dicedpixels: Yep, `KeyBindingHelper`. We'll have to live with it till the Mojmap renames are done.

reference/latest/src/client/java/com/example/docs/keymapping/ExampleModKeyMappingsClient.java
its-miroma: imo that should be mentioned in the article, as well as a link to the standard US layout. maybe either:
https://kbdlayout.info/kbdus
https://en.wikipedia.org/wiki/QWERTY#/media/File:KB_United_States.svg

dicedpixels: I feel like that would overcomplicate things for a new user 🤔. Maybe let's see what others think?

its-miroma: I don't think it would if worded carefully, something like:

```md
::: info

Note that the names of the key tokens (`GLFW.GLFW_KEY_*`) assume a [standard US layout](...).

That means, for example, that if you're using an AZERTY layout, pressing on A would yield `GLFW.GLFW_KEY_Q`.

:::
```

what do others think?

Fellteros: I think it's good to mention it, maybe it would be beneficial to mention if it's QWERTY or QWERTZ (I don't know how much these two are used)

dicedpixels: I've now added the info box Miroma suggested. I think it's simple enough to not confuse people with GLFW key/scan codes.

develop/key-mappings.md
```diff
@@ -0,0 +1,71 @@
...
+Many of these key mappings can be configured through the settings menu.
+
+With help of Fabric API, you can create your own custom key mappings and react to them in your mod.
+
+Key mappings are exclusive to the client. So, registration and reacting to key mappings should
+be done on the client side. You can use the **client initializer** for this.
+
+## Creating a Key Mapping {#creating-a-key-mapping}
+
+A key mapping consist of two parts. The mapping to a key and the category it belongs to.
```
cassiancc:
```suggestion
A key mapping consists of two parts: the mapping to a key, and the category it belongs to.
```

develop/key-mappings.md
```diff
@@ -0,0 +1,71 @@
...
+  - cassiancc
+  - dicedpixels
+---
+
+Minecraft handles user input from peripherals such as the keyboard and mouse using key mappings.
+Many of these key mappings can be configured through the settings menu.
+
+With help of Fabric API, you can create your own custom key mappings and react to them in your mod.
+
+Key mappings are exclusive to the client. So, registration and reacting to key mappings should
```
cassiancc:
```suggestion
Key mappings only exist on the client side. This means that registration and reacting to key mappings should
```

develop/key-mappings.md
```diff
@@ -0,0 +1,71 @@
...
+Next, we can create a key mapping. We will be using Fabric API's `KeyBindingHelper` to register
+our key mapping at the same time.
+
+@[code lang=java transcludeWith=:::key_mapping](@/reference/latest/src/client/java/com/example/docs/keymapping/ExampleModKeyMappingsClient.java)
+
+::: info
+
+Note that the names of the key tokens (`GLFW.GLFW_KEY_*`) assume a [standard US layout](https://upload.wikimedia.org/wikipedia/commons/d/da/KB_United_States.svg).
+
+That means, for example, that if you're using an AZERTY layout, pressing on A would yield `GLFW.GLFW_KEY_Q`.
```
cassiancc:
```suggestion
This means that if you're using an AZERTY layout, pressing on <kbd>A</kbd> would yield `GLFW.GLFW_KEY_Q`.
```

develop/key-mappings.md
```diff
@@ -0,0 +1,71 @@
...
+
+:::
+
+Sticky keys can also be created with `KeyBindingHelper` by passing a `ToggleKeyMapping` instance.
```
cassiancc:
```suggestion
Sticky keys can also be created with `KeyBindingHelper` by passing a `ToggleKeyMapping` instance instead of a `KeyMapping`.
```

develop/key-mappings.md
```diff
@@ -0,0 +1,71 @@
+---
+title: Custom Key Mappings
```
cassiancc: Why "Custom Key Mappings" rather than simply "Key Mappings"?

cassiancc (changes requested):

develop/key-mappings.md
```diff
@@ -0,0 +1,71 @@
...
+
+![Translated Key Category and Mapping](/assets/develop/key-mappings/translated.png)
+
+## Reacting to Key Mappings {#reacting-to-key-mappings}
+
+Now that we have a key mapping, we can react to it using a client tick event.
+
+@[code lang=java transcludeWith=:::client_tick_event](@/reference/latest/src/client/java/com/example/docs/keymapping/ExampleModKeyMappingsClient.java)
+
+This will print "Key Pressed!" to the in-game chat every time the mapped key is pressed.
```
its-miroma:
```suggestion
This will print "Key Pressed!" to the in-game chat every time the mapped key is pressed. Keep in mind that holding the key will repeatedly print the message to the chat, so you might want to implement guards so this logic only triggers once.
``` 

to avoid repeating "logic"

dicedpixels: I'd argue that the repeating logic could the desired behavior. There could be a valid use case for repeated execution, like a gun or something.

develop/key-mappings.md
dicedpixels: Took it off the Wiki. We can indeed go with "Key Mappings".

cassiancc (dismissed):

develop/key-mappings.md
cassiancc: It could be, but it bears mentioning in case the intending use case is something that shouldn't be repeatedly triggered like a config change or opening a screen.

its-miroma (approved):

its-miroma (changes requested):
Wait no, there's still an unsatisfied suggestion by @cassiancc

develop/key-mappings.md
dicedpixels: Added a slightly different version of this suggestion. I hope this conveys the idea that handling repeated presses is up to the developer.

cassiancc (approved):
Looks good, my changes are addressed.

its-miroma (approved):

PR #439, opened 2025-12-27
update CI workflows

its-miroma:
Please read through the commit messages, which contain important context and summary


.github/workflows/i18n-pull.yaml
```diff
@@ -0,0 +1,76 @@
...
+name: Pull from Crowdin
+
+on:
+  schedule:
+    - cron: 37 7 * * SAT
+  workflow_dispatch:
+
+concurrency:
+  group: i18n-pull
+  cancel-in-progress: true
```
modmuss50: Nit: I dont think this is wise for an action that alters stuff on the repo.

.github/workflows/i18n-push.yaml
```diff
@@ -0,0 +1,50 @@
...
+      - "*_translations.json"
+      - "crowdin.yml"
+      - "!README.md"
+      - "!translated/**"
+      - "!versions/**"
+  workflow_dispatch:
+
+concurrency:
+  group: i18n-push
+  cancel-in-progress: true
```
modmuss50: Again, doesnt sound right to me. Should be false.

modmuss50 (commented):
Just some minor points, we should ideally wait for MR to confirm the license.

its-miroma:
Modrinth have confirmed their workflows are under MIT, so this is ready for review @modmuss50

modmuss50 (dismissed):
Ping me on discord when you want to merge this.

PR #445, opened 2026-01-01
use FontSource, fix lang attribute

its-miroma:
I'd appreciate feedback on font choices. I chose JetBrains Mono which afaik is used in IntelliJ. I also used Noto Color Emoji for flags. I did not track the fonts in the repo, rather I used FontSource per @modmuss50's suggestion.


cassiancc:
Font changes are pretty invisible to my eyes, even switching side by side, so no issues there for me. I definitely would not want serifs in the body, though I wouldn't be opposed to testing how it looks in a header. As long as there's fallback fonts, other language support shouldn't be an issue as far as I know.

.vitepress/theme/style.css
```diff
@@ -3,15 +3,8 @@
...
-@font-face {
-  font-family: "Flag Emoji Polyfill";
-  unicode-range:
-    U+1F1E6-1F1FF, U+1F3F4, U+E0062-E0063, U+E0065, U+E0067, U+E006C, U+E006E, U+E0073-E0074,
-    U+E0077, U+E007F;
-  src: url("./CountryFlagsPolyfill.woff2") format("woff2");
-  font-display: swap;
-}
+/* Google Fonts: JetBrains Mono, Noto Color Emoji (Windows does not render flags) */
+@import url("https://fonts.googleapis.com/css2?family=JetBrains+Mono:ital,wght@0,100..800;1,100..800&family=Noto+Color+Emoji&display=swap");
```
modmuss50: Is it possible to not have any external CDN dependencies? Im not strongly against this, it just feels nicer not to have them.

modmuss50: I have just found: https://fontsource.org/docs/getting-started/introduction It has both these fonts in a way that we should be able to easily depend on via npm. This is much better than including them in the repo or pulling them from google.

its-miroma: Reading into it, Google Fonts doesn't seem to be reliable long-term, as breaking changes get introduced without opt-out.
see: https://github.com/google/fonts/issues/1307

I'll take a look at fontsource

modmuss50 (approved):

PR #463, opened 2026-01-13
fix: use correct locale path in 404 page

PikachuTW:
## Description

On non-English pages, the href for "Take me home" button is incorrect.

The expected path is `/zh_tw/`, but it gives `/zh-tw/`.

Additionally, the "read in English" button is also broken, which links to the current page instead of the English version.

## Fixes

This occurs because `data.lang.value` returns `zh-tw`, while the route path use `zh_tw`. To fix this, I replaced it with `data.localeIndex.value`.

This also resolve the second issue by making the regex replacement with the correct string.


.vitepress/theme/components/NotFoundComponent.vue
```diff
@@ -8,7 +8,7 @@ import { Fabric } from "../../types";
 const data = useData();
 
 const options = computed(() => data.theme.value.notFound as Fabric.NotFoundOptions);
-const removeForEnglishRegex = new RegExp(String.raw`^${data.lang.value}/|\.md$`, "g");
+const removeForEnglishRegex = new RegExp(String.raw`^${data.localeIndex.value}/|\.md$`, "g");
 
 const urls = computed(() =>
   data.lang.value === "en-us"
```
its-miroma:
```suggestion
  data.localeIndex.value === "root"
```
for uniformity

its-miroma (changes requested):
Thank you!

.vitepress/theme/components/NotFoundComponent.vue
PikachuTW: Thanks! I've updated it.

its-miroma (approved):

cassiancc (approved):

PR #474, opened 2026-01-23
Fix the enchantment generator register snippet

CelDaemon:
Uses a transclude instead of line ranges to prevent issues due to shifting in example code.


CelDaemon:
Seems like this causes issues with other transcludes, I'll change those to use markers too.

its-miroma (dismissed):

its-miroma (dismissed):

its-miroma (approved):

PR #488, opened 2026-02-03
Add available 26.1 documentation

cassiancc:
While still not exhaustive as the version isn't out yet, we should be able to include the available docs for 26.1 porting on its page.


its-miroma (approved):
imo, this page should have been under /26.1/porting, instead of /porting/next, however that would require some careful testing of the framework and redirects would be helpful. i'll keep this in mind

PR #491, opened 2026-02-10
Added a description of a known issue

BuelieR:
Added a description of a known issue.
Using the official mapping makes it impossible to instantiate Identifier, but most tutorials nowadays are based on the YARN mapping and do not explain this in detail.

### DIFF
--- a/develop/items/first-item.md
+++ b/develop/items/first-item.md
@@ -42,6 +42,8 @@ This will not work if you've marked the item as damageable, as the stack size is
 
 @[code transcludeWith=:::2](@/reference/latest/src/main/java/com/example/docs/item/ModItems.java)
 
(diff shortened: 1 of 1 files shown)


its-miroma:
What about `Identifier.fromNamespaceAndPath()`?

ekulxam:
`Identifier` being abstract is not caused by mojmap. Mappings only change the name of the class, not the access modifiers or the function of the class. Additionally, the modern fabric docs are in mojmap and correctly use `Identifier.fromNamespaceAndPath`, which is `Identifier.of` in yarn. Notice how both methods perform the same function (although they have different names due to mappings), and Identifier is still abstract.

its-miroma:
> most tutorials nowadays are based on the YARN mapping

while that might have been true back in 2024, the docs now are fully migrated to Mojmap

PR #492, opened 2026-02-10
Fix #363

### LINKED ISSUE #363: "Custom Data Components": `appendTooltip` is deprecated
In [this section](https://github.com/FabricMC/fabric-docs/blob/45413c66b010c8ad099916aafbde3b3581bdffd4/develop/items/custom-data-components.md?plain=1#L77) it mentions you need to implement `appendTooltip`, but since 1.21.5 that's deprecated and even on the old wiki it says [you need to use Fabric Api](https://wiki.fabricmc.net/tutorial:tooltip). Consider changing this section.

ekulxam:


cassiancc (approved):
This works for now. I'm of the opinion the article should likely be rewritten to introduce custom component classes earlier, but that can likely wait until if `appendHoverText` stops functioning.

PR #493, opened 2026-02-10
Rename translated/ru_ru/players/installing-fabric/windows.md to trans…

siciksasa-creator:
…lated/ru

Ееенеее


modmuss50:
Im not sure what the goal here is, but it doesnt seem correct.

its-miroma:
I think @siciksasa-creator is trying to apply point 7 of #444. Please follow progress along there

PR #504, opened 2026-02-20
Use debug log level over development environment check

### LINKED ISSUE #356: "Debugging Mods": use launch config instead of `devLogger`
#330 is great, and I'm happy it's been merged. ~~Now I can bike shed it without it being blocked!~~
The `isDevelopmentEnvironment` check for logging gets kind of cumbersome - I prefer to use `LOGGER.debug` and configure launch configs to enable debug logging. It should be fairly easy to make the change, and I'm willing to PR it if others agree. Is this something that would be wanted?

CelDaemon:
Adds a log4j config that enables debug logging, and encourages its use for any development-specific logging.

Fixes #356


develop/debugging.md
```diff
@@ -58,11 +59,39 @@ In the logged line, you can find:
...
+
+### Debug Logging {#debug-logging}
+
+The debug log level is hidden by default, but can be specifically enabled in the development environment using some configuration.
+
+To do this, create the `log4j-dev.xml` configuration file in your project's root directory, and add the following contents. Make sure to replace the value inside the `name` attribute with your mod's modid.
+
+@[code lang=xml](@/reference/latest/log4j-dev.xml)
+
+Then in the `build.gradle`, tell loom to use our new configuration.
```
cassiancc:
```suggestion
Then in the `build.gradle`, tell Loom to use our new configuration.
```

reference/latest/log4j-dev.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Configuration>
+  <Loggers>
+    <Logger name="example-mod" level="debug" additivity="false">
+      <AppenderRef ref="SysOut"/>
+      <AppenderRef ref="ServerGuiConsole"/>
+    </Logger>
+  </Loggers>
+</Configuration>
```
modmuss50, line 9: Just set `-Dfabric.log.level=debug`. No need for a custom config like this.

reference/latest/src/main/java/com/example/docs/debug/ExampleModDebug.java
```diff
@@ -26,25 +25,17 @@ public class ExampleModDebug implements ModInitializer {
...
 				new TestItem(new Item.Properties().setId(testItemKey).stacksTo(1).rarity(Rarity.EPIC)
 						.component(DataComponents.CUSTOM_NAME, Component.literal("[Use on Stone Block]"))));
 	}
 
-	// :::problems:dev-logger
-	// This method will only log if the Minecraft instance
-	// is running in a Development Environment, like your IDE
-	public static void devLogger(String loggerInput) {
-		if (!FabricLoader.getInstance().isDevelopmentEnvironment()) {
-			return;
```
modmuss50, line 41: Good, this should go its clearly wrong 👍

develop/debugging.md
```diff
@@ -58,11 +59,39 @@ In the logged line, you can find:
...
+
+@[code lang=java transcludeWith=:::debug-logging](@/reference/latest/build.gradle)
+
+We can now send debug log messages to the console.
```
cassiancc:
```suggestion
Now, when we send debug log messages, we'll be able to see them in the console.
```

develop/debugging.md
```diff
@@ -58,11 +59,39 @@ In the logged line, you can find:
...
+
+:::
+
+If you are still unsure which log level to choose, a good rule of thumb is to only use non-debug levels if something went wrong. Modpack devs and users don't care too much about, for example, items initializing; they would rather know if, for example, a datapack failed to load correctly.
+
+### Debug Logging {#debug-logging}
+
+The debug log level is hidden by default, but can be specifically enabled in the development environment using some configuration.
+
+To do this, create the `log4j-dev.xml` configuration file in your project's root directory, and add the following contents. Make sure to replace the value inside the `name` attribute with your mod's modid.
```
cassiancc:
```suggestion
To do this, create the `log4j-dev.xml` configuration file in your project's root directory, and add the following contents. Make sure to replace the value inside the `name` attribute with your mod's modid. This configuration allows debug logs from your mod to be shown in the console.
```

cassiancc (changes requested):
Looks good! A few minor revisions.

reference/latest/log4j-dev.xml
CelDaemon: That enables debug logs for all loggers, resulting in a lot of log spam. This config only enables it for the mod itself.

modmuss50: Specifying it based on the logger name isnt ideal but might be ok for this. Ideally you should set the logger name to be that of the class, and then use a log marker. Its not something I have ever done before tho.

CelDaemon: I've also proposed that before, but it seems like people generally prefer the logger name to be the mod ID. With the class name, it's sometimes less clear which mod a log entry comes from.

I could still try looking into markers, but thought it would be simpler like this.

reference/latest/src/main/java/com/example/docs/debug/ExampleModDebug.java
CelDaemon: Yeah, it also encourages people to use raw strings instead of parameterised logging. It might still be useful to show off the possibility of checking for the development environment, but I didn't think it to be worth the trouble.

reference/latest/log4j-dev.xml
modmuss50: Lets keep it simple and go with this. Its soo much better than before. I wonder if there would be scope for loom to help with this in the future.

CelDaemon: That'd be very cool! I did think about trying something with loom too as the base config is provided by it as well, but I'm not sure how to handle cases where the loggers use class names Vs the mod ID convention.

cassiancc (approved):

modmuss50 (approved):
This is a good improvement, if we ever add a better way to do this in the future its easy enough to update.

Issue #499, opened 2026-02-18
Yarn Mappings Are Not Supported Anymore


arqyt:
I am developing an arg mod and am using yarn mappings. The problem is that I cannot seem to find where the docs use yarn mappings instead of Mojang mappings. Is there a version of the docs that uses yarn mappings?

cassiancc:
Yarn Mappings are [no longer supported by Fabric](https://fabricmc.net/2025/10/31/obfuscation.html#the-future-of-yarn) as part of preparation for 26.1 and above. Please [migrate to Mojang Mappings](https://docs.fabricmc.net/develop/migrating-mappings/), the process is automated.

arqyt:
I really don't want to have to learn different mappings. I only have used yarn for all of my mods.

PR #508, opened 2026-02-22
Update ModItems to use new Item.Settings class

gi199:


cassiancc:
`Item.Settings` is incorrect, that's a yarn name, see #499 .

PR #510, opened 2026-02-22
Rename `GenericItem` to `T`

CelDaemon:
This makes it clearer that GenericItem is a generic type parameter, and not a class name. Many people directly copy this code and get confused by the naming.

### DIFF
--- a/develop/items/first-item.md
+++ b/develop/items/first-item.md
@@ -24,7 +24,7 @@ Mojang does this with their items as well! Check out the `Items` class for inspi
 
 @[code transcludeWith=:::1](@/reference/latest/src/main/java/com/example/docs/item/ModItems.java)
 
-Notice how we're using `T`, which allows us to use the same method `register` for registering any type of item that extends `Item`. We're also using a [`Function`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/function/Function.html) interface for the factory, which allows us to specify how we want our item to be created given the item properties.
+Notice how we're using `T`, a [generic type](https://docs.oracle.com/javase/tutorial/java/generics/types.html) that extends `Item`. This allows us to use the same method `register` for registering any type of item that extends `Item`. We're also using a [`Function`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/function/Function.html) for the factory, which allows us to specify how we want our item to be created given the item properties.
 
 ## Registering an Item {#registering-an-item}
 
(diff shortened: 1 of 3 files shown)


develop/items/first-item.md
```diff
@@ -24,7 +24,7 @@ Mojang does this with their items as well! Check out the `Items` class for inspi
 
 @[code transcludeWith=:::1](@/reference/latest/src/main/java/com/example/docs/item/ModItems.java)
 
-Notice how we're using a `GenericItem`, which allows us to use the same method `register` for registering any type of item that extends `Item`. We're also using a [`Function`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/function/Function.html) interface for the factory, which allows us to specify how we want our item to be created given the item properties.
+Notice how we're using `T`, which allows us to use the same method `register` for registering any type of item that extends `Item`. We're also using a [`Function`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/function/Function.html) interface for the factory, which allows us to specify how we want our item to be created given the item properties.
```
dicedpixels: Can we add something like `T`, a generic type parameter, which allows...

I don't think we need to explain what a generic type parameter is. You're expected to have an understanding of generics.

dicedpixels (changes requested):

develop/items/first-item.md
cassiancc: Linking to an explanation on generics would be fine, we do the same for functions later on.

its-miroma:
```suggestion
Notice how we're using `T`, a [generic type](https://docs.oracle.com/javase/tutorial/java/generics/types.html) that extends `Item`. This allows us to use the same method `register` for registering any type of item that extends `Item`. We're also using a [`Function`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/function/Function.html) for the factory, which allows us to specify how we want our item to be created given the item properties.
```

dicedpixels: I think it's better to drop the word "interface". `Function` is a functional interface. Otherwise it's good.

Earthcomputer:
I think a lot of us aren't a big fan of the convention to use single-letter type names in Java for generic types, but this is a convention that comes from Java which Mojang also uses and there's not a lot we can do to change it and anything other than this convention is not what people are expecting. Perhaps using `I` for "item" is the most we could do.

modmuss50 (approved):
I think T is fine. We could bikeshead this all day long if we wanted to.

cassiancc (approved):

its-miroma (approved):

dicedpixels (approved):

PR #512, opened 2026-02-27
fix some things in framework

its-miroma:
- don't force posix mode which breaks on windows
- don't add and hide the banner in prod
- don't add meta tags to 404 pages


dicedpixels (approved):

MildestToucan:
Won't review because I have practically no literacy in the type of technical changes being made here, but I will comment for the sake of record-keeping that I'm the one who noticed the issue with `posix` on Windows, and that the change away from `posix` from Miroma did appear to fix my issue when I did it locally.

PR #516, opened 2026-03-01
Fix crash when opening inventory

CelDaemon:
Fixes the exception thrown by the example tooltip callback. This exception occurs when handling the tooltip of any item that does not have the counter component. Due to a `get` returning a null value, unboxing to an `int` primitive fails instead of being handled gracefully.

### DIFF
--- a/reference/latest/src/client/java/com/example/docs/ExampleModClient.java
+++ b/reference/latest/src/client/java/com/example/docs/ExampleModClient.java
@@ -24,11 +24,9 @@ public class ExampleModClient implements ClientModInitializer {
 		ItemTooltipCallback.EVENT.register((stack, context, type, tooltip) -> {
 			Integer count = stack.get(ModComponents.CLICK_COUNT_COMPONENT);
 
-			if (count == null) {
-				return;
+			if (count != null) {
+				tooltip.add(Component.translatable("item.example-mod.counter.info", count).withStyle(ChatFormatting.GOLD));
 			}
-
-			tooltip.add(Component.translatable("item.example-mod.counter.info", count).withStyle(ChatFormatting.GOLD));
 		});
 		// #tooltip_provider_client
 	}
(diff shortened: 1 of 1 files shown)


its-miroma (dismissed):
if the transclusions aren't broken, this LGTM.
also, please double check the article to see if it still matches up

CelDaemon:
image
Seems to be fine! I also checked for any line-based transclusions using this class, and couldn't find any.

I'm kinda wondering if I should invert the if statement though.

reference/latest/src/client/java/com/example/docs/ExampleModClient.java
```diff
@@ -22,8 +22,11 @@ public void onInitializeClient() {
 
 		// #tooltip_provider_client
 		ItemTooltipCallback.EVENT.register((stack, context, type, tooltip) -> {
-			int count = stack.get(ModComponents.CLICK_COUNT_COMPONENT);
-			tooltip.add(Component.translatable("item.example-mod.counter.info", count).withStyle(ChatFormatting.GOLD));
+			Integer count = stack.get(ModComponents.CLICK_COUNT_COMPONENT);
+
+			if (count != null) {
+				tooltip.add(Component.translatable("item.example-mod.counter.info", count).withStyle(ChatFormatting.GOLD));
```
its-miroma, line 28: What happens if the component is removed from the item? Does the previous tooltip disappear automatically?

cassiancc: Yes, `ItemTooltipCallback` is called every tick.

CelDaemon: Yes, the event gets invoked when the game wants to refresh the item's tooltip with entirely new lines. The event adds to a fresh list every time, so simply not adding the component effectively removes it from the tooltip.

its-miroma (approved):

cassiancc (approved):

Issue #415, opened 2025-11-28
Transclusions Should Drop `:::` Comments


its-miroma:
![Example](https://github.com/user-attachments/assets/02e716b1-4408-4fc3-bd34-c2f6f21d8c7c)

| Reported by Whom |     When     |                                                 Where                                                 |
| ---------------- | :----------: | :---------------------------------------------------------------------------------------------------: |
| @JR1811          | `2024-02-28` | [on Discord](https://discord.com/channels/507304429255393322/1208846408552030238/1212268520956498000) |
| @skycatminepokie | `2024-12-24` | [on Discord](https://discord.com/channels/507304429255393322/1208846408552030238/1320889968993763339) |
| @JR1811 (again)  | `2025-05-22` | [on GitHub](https://github.com/FabricMC/fabric-docs/pull/246#discussion_r2102987551)         |
| @dicedpixels     | `2025-11-21` | [on Discord](https://discord.com/channels/507304429255393322/1208846408552030238/1441488997737037845) |
| @cassiancc       | `2025-11-21` | [on Discord](https://discord.com/channels/507304429255393322/1208846408552030238/1441489531277807626) |
| @ekulxam         | `2026-01-23` | [on Discord](https://discord.com/channels/507304429255393322/1208846408552030238/1464183184953704458) |
| @MildestToucan   | `2026-02-28` | [on Discord](https://discord.com/channels/507304429255393322/1208846408552030238/1477286856549728277) |
| @CelDaemon       | `2026-02-10` | [on Discord](https://discord.com/channels/507304429255393322/1208846408552030238/1470711032560357581) |
| @Earthcomputer   | `2026-03-09` | [on Discord](https://discord.com/channels/507304429255393322/1208846408552030238/1480618769813340211) |
| @Wind292         | `2026-03-15` |         [on GitHub](https://github.com/FabricMC/fabric-docs/pull/528#issuecomment-4063814056)         |
| @JaaiDead        | `2026-03-21` | [on Discord](https://discord.com/channels/507304429255393322/1208846408552030238/1484930661260460163) |

PR #528, opened 2026-03-15
Inconsistent trailing newlines in codeblocks

Wind292:
I noticed that some of the codeblocks have a trailing newline while most others don't. This extra newline also makes the codeblock slightly scrollable, which can be annoying if your mouse happens to be on a codeblock. I applied a fix on the items category, should I go through and fix all of them? or is this intentional for some reason?

See the third codeblock at https://docs.fabricmc.net/develop/items/first-item. There are many more besides this one example. Notice the scroll bar and the last line having a line number but it just being an empty newline.

image

after fix 

image


package-lock.json
its-miroma, line 1: these changes are unexpected. What version of `npm` are you on?

its-miroma (commented):
Thank you, this is a good idea! It was like this because there was a conflicting checkstyle rule, but [that was changed recently](https://github.com/FabricMC/fabric-docs/blame/00c07d7295ee3612f2183ca36abf09c0551e2471/reference/checkstyle.xml#L58-L59) and wasn't applied everywhere I guess. Maybe using a regex to look for `\}\n\n\t*\/\/` will suffice?

package-lock.json
Wind292: Version 11.10.1, on Arch Linux [btw]

Wind292:
>  Maybe using a regex to look for `\}\n\n\t*\/\/` will suffice?

I don't think just that check will work (it might *help* though; catching some cases). Sometimes people use the `// #entrypoint` with it ending at the end of the file, and because the linter wants a newline at the end of every file this would guarantee a newline, (there are other cases besides this one). Is there a way to check the vitepress's codeblocks to see if it has an extra newline displayed in the website?

Wind292:
I'm also noticing that a few of the codeblocks are displaying `// ::: foo-bar`. This is also very confusing and should be fixed.
image

cputnam-a11y:
> I'm also noticing that a few of the codeblocks are displaying `// ::: foo-bar`. This is also very confusing and should be fixed. image

This one is waiting on an upstream pr iirc

develop/data-generation/advancements.md
```diff
@@ -117,7 +117,10 @@ Almost there! Next, we need an instance of our criterion to work with. Let's put
 
 To make sure that our criteria are initialized at the right time, add a blank `init` method:
 
-@[code lang=java transcludeWith=:::datagen-advancements:mod-criteria-init](@/reference/latest/src/main/java/com/example/docs/advancement/ModCriteria.java)
+```java
```
its-miroma: No, please don't do this. I know it's ugly, but that's a known issue #415
Ugly transclusion comments are a temporary compromise (hopefully vuejs/vitepress#5014)

develop/data-generation/advancements.md
```diff
@@ -155,7 +158,7 @@ Moving on, we now need to fix our `trigger` method:
 
 If you've made a new criterion, we need to add it to `ModCriteria`
 
-@[code lang=java transcludeWith=:::datagen-advancements:new-mod-criteria](@/reference/latest/src/main/java/com/example/docs/advancement/ModCriteria.java)
+@[code lang=java transclude={11-14}](@/reference/latest/src/main/java/com/example/docs/advancement/ModCriteria.java)
```
its-miroma: Don't do this either please. from experience, line-number-based transclusions are too brittle

its-miroma (changes requested):
Do please revert the changes to `package-lock.json`. You should notice that running `npm i` again will do just that.

its-miroma:
Sorry, I was reviewing just as you pushed a commit reverting (:

its-miroma (approved):

reference/latest/src/main/java/com/example/docs/block/custom/PrismarineLampBlock.java
```diff
@@ -62,8 +62,8 @@ public static int getLuminance(BlockState currentBlockState) {
 		// Return a light level if activated = true
 		return activated ? 15 : 0;
 	}
-
 	// :::5
+
```
its-miroma: remove trailing whitespace

reference/latest/src/main/java/com/example/docs/block/ModBlocks.java
```diff
@@ -64,6 +64,7 @@ public class ModBlocks {
 			true
 	);
 	// :::4
+
```
its-miroma: remove trailing whitespace

its-miroma (changes requested):

Wind292:
Okay I think that's all. I'm going to undraft this now

reference/latest/src/main/java/com/example/docs/item/custom/CustomSoundItem.java
```diff
@@ -27,8 +27,8 @@ public InteractionResult interactLivingEntity(ItemStack stack, Player user, Livi
 
 		return super.interactLivingEntity(stack, user, entity, hand);
 	}
-
 	// :::1
+
```
its-miroma: remove trailing whitespace

its-miroma (changes requested):

cassiancc (approved):

its-miroma (approved):

PR #529, opened 2026-03-16
fix: apply vuejs/vitepress#5014

its-miroma:
I checkedout vitepress@v2.0.0-alpha.12, cherry-picked the changes from vuejs/vitepress#5014 (except the additional configuration option `stripMarkersFromSnippets`), built the resulting source tree, then replaced `node_modules/vitepress/dist` with my build, then ran `pnpm patch vitepress` to generate this and `pnpm patch-commit`.

Another future PR will run an automated migration, thereby fixing #415


CelDaemon:
That doesn't seem right? There's lots of changes in there unrelated to the transclusion PR.

its-miroma:
I don't really know tbh... That's why it's a draft

CelDaemon:
I could give it a shot to generate a cleaner patch.

its-miroma:
@CelDaemon @Earthcomputer why is it not working now?!

CelDaemon:
Oh that's much better!

Earthcomputer:
Yeah looks good

CelDaemon:
Hmmm, patch does indeed look better but doesn't apply correctly.

CelDaemon:
I think the problem may be that your build resulted in a different chunk filename, maybe this can be fixed manually?

Earthcomputer:
What's the best page to look at so we can see the effect of this change?

its-miroma:
@Earthcomputer No page takes advantage of this PR yet.
You could apply one of the changes from this commit locally to see the effects:
https://github.com/its-miroma/fmc/commit/a76d6366ff47f957e300079a57c872c2d9428af9

---

1. Change `@[code transcludeWith=:::REGION-NAME](@/path/to/file.java)` to `<<< @/path/to/file.java#REGION-NAME`
2. In `@/path/to/file.java`, change all `// :::REGION-NAME` to `// #region REGION-NAME` or `// #endregion REGION-NAME`

Notice how, in the dev preview, when modifying `@/path/to/file.java` the snippet will update.

its-miroma:
@Earthcomputer I just pushed the result of the migration of `block-containers`. This is the difference:

| [Before](https://docs.fabricmc.net/develop/blocks/block-containers#:~:text=int%20timeSinceDropped) | [After](https://deploy-preview-529--nimble-elf-d9d491.netlify.app/develop/blocks/block-containers#:~:text=int%20timeSinceDropped) |
|--------|-------|
| before | after |

IMB11:
Ooi, it's probably cleaner to switch to pnpm and use the built in `pnpm patch` - just a suggeston though https://pnpm.io/cli/patch

its-miroma:
@IMB11 100% agree, especially as pnpm supports `minimumreleaseage` which @modmuss50 expressed interest in

<https://discordapp.com/channels/507304429255393322/1208846408552030238/1465051405877514405> via <https://discord.fabricmc.net/>

modmuss50 (approved):

PR #553, opened 2026-04-01
chore: address any NPM vulnerability

its-miroma:
> [!WARNING]
> **HAPPY APRIL FOOLS!**

We've all seen the situation with NPM deteriorating, and yesterday we literally lived through ["the biggest hack of 2026" (for now)](https://www.youtube.com/watch?v=yiLIZLPNEm8) just a day ago.

After some internal discussion between @its-miroma and myself, we ended up realizing that the only way out of this is to go back to square one. It has become clear that adding all these levels of abstraction (VitePress, Vue, NPM, etc.) is just adding additional potential breaking points for free.

The solution is clear: this PR simply avoids the building step of converting MD files to servable HTML (which is where NPM would've come in), because we realized we can just track the deployed files directly. Editing HTML isn't that difficult, and the graphical UI doesn't need any more changes because it's clearly perfect already.

This PR also conveniently drops all other unrelated files: we don't need Crowdin (translators can just translate the HTML), we don't need the Reference Mod (code can be placed within the HTML), and therefore we can also finally drop the snippet plugin and solve that issue (FINALLY).

The only issue I can see is that, because this huge PR changes every single file in the repo, we'll need every contributor and every author to approve it before we can merge it. Please get to work everyone.


its-miroma:
Thank you Miroma for opening this PR. Yeah, after discussion we had yesterday this was the only sensible and rational solution. :+1:

its-miroma:
@Copilot let me briefly tackle your suggestions:

1. Regarding the title of the PR, addressing the NPM vulnerabilities is the most important thing this PR does. Now, the fact that every single file on the repo has been modified is a sad consequence but for the greater good
2. I know I removed all CI workflows, and that's because GitHub Actions is arguably even more vulnerable than NPM. We won't need any checks, because we'll directly deploy the website on GitHub Pages from the branch
3. Yea, redirects unfortunately might break, although I tried to preserve some in `404.html`. However, this problem cannot be solved because someone (ehmm *cough* GitHub Pages) does not support 301 redirects in the big '26 :wilted_flower:
4. `.editorconfig` is no longer needed for HTML files, because browser will accept anything we throw at them even when it's not formatted
5. `.gitattributes` is no longer needed because there are no longer any generated files in the repo—everything is already within the HTML

its-miroma:
@modmuss50 it seems that Copilot thinks everything's good (the low-confidence comments aren't actually problems as you can see from the message above), so it approves :+1:

dicedpixels (approved):
Image

CelDaemon (approved):
I think copilot is just speechless at how quickly and efficiently we have handled the supply chain issues.

cassiancc:
Happy April fools, all.

PR #564, opened 2026-04-06
Biome and Dimension Generation Page

### LINKED ISSUE #563: Developer Guides: Biome Data Generation
Biome datagen is missing, along with most other worldgen-datagen pages like dimensions etc. I can open a PR for a biome-datagen page if that is something that is wanted.

Wind292:
also fixed a small bug in the features page where the code was referenced by line instead of using the ::: reference, I initially authored that page, it was *my* bad code. 

Also a page for biome spawning still needs to be implemented, this page was only for biome creation.

Fixes #563


Wind292:
What is the proper implementation for making biomes naturally generate?  I believe the best practice is Terrablender, or does fabric have a built in biome generation.

cassiancc:
I'm not sure if there's a built-in option. My usual recommendation to other devs is [Biolith](https://modrinth.com/mod/biolith), it's maintained by the developers of Mod Menu and seems to be more used than Terrablender these days.

NotNightSky:
@cassiancc I am thinking about taking up the work of biome generation. So should I write about biome generation using the Biolith api or mixins and stuff

cputnam-a11y:
Would it make sense to defer this, and simply add the biome to a modded dimension?

NotNightSky:
Yes, that would be the simplest option

NotNightSky:
I was writing data generation files for a custom dimension and I am getting this error, `java.util.concurrent.CompletionException: java.lang.IllegalArgumentException: Registry minecraft:dimension is not loaded from datapacks`. which according to me does not make a whole lot of sense as dimensions are added by using datapacks

NotNightSky:
I am unable to continue due to a bug in the [Fabric Api](https://github.com/FabricMC/fabric-api/issues/3838) which causes the error `java.util.concurrent.CompletionException: java.lang.IllegalArgumentException: Registry minecraft:dimension is not loaded from datapacks`. let me see if I can find a work around

cputnam-a11y:
The issue should describe the current best options

NotNightSky:
I have completed the dimension creation page by hard coding the dimension json as a work around to the registry error (DimensionType is generated via datagen) and I am testing stuff out for biome generation

NotNightSky:
@Wind292 could you possibly visit the PR that I made to the draft branch

Wind292:
> @Wind292 could you possibly visit the PR that I made to the draft branch

Just did it. Totally forgot to merge thanks for the reminder

develop/data-generation/dimension-creation.md
```diff
@@ -0,0 +1,162 @@
...
+### Teleporting to the Dimension {#dimensiontp}
+
+Currently, we don't have any portal block or item or event setup to teleport us to the dimension. Therefore, to teleport to the dimension we will have to use this command:
+
+`/execute in <mod-id>:<dimension> run tp <username> ~ ~ ~`
+
+::: info INFO
+
+the command in this guide will be:
+`/execute in example-mod:example_dimension run tp Player986 ~ ~ ~`
```
Wind292, line 155: Do we want to phrase this such that `Player986` is replaced with `@p`? Or `PlayerXXX`.

CelDaemon: Why not just use `@s`? I'm not even sure `@p` will work considering the dimension context has changed.

Wind292: No you're right `@p` won't work. It would have to be `@s`. Though do we want this changed? With `@s` the user can copy and paste it. But with a player name it shows the user how the command works. But this is fabric docs not Minecraft command docs so I'm leaning towards `@s`

develop/data-generation/dimension-creation.md
```diff
@@ -0,0 +1,162 @@
...
+
+Here is the preview of the example world:
+
+![Dimension Preview](/assets/develop/data-generation/dimension-generation/ingame_preview.png)
+
+:::
+
```
its-miroma, line 162: this fixes CI:

```suggestion

<!---->

```

NotNightSky:
Does anyone have a way to add biomes to the biome pools of existing dimensions, or should we use Biolith as suggested by cassiancc (the usage of which I personally don't recommend based on the fact that we are writing documentation)?

develop/data-generation/dimension-creation.md
NotNightSky: As a guy using Fedora, I lean towards `Player986` because I think that people should know how the command works before using it.

Wind292:
AFAIK There isn't a clean way to add biomes to the overworld without a third party mod. You can overwrite the overworld's json file to insert a custom biome, but compatibility with other mods is basically impossible. I don't know what the best solution here is. Though I believe adding new biomes to the nether is possible, for what its worth.

cputnam-a11y:
The recommended solution here is to use a 3rd party library. I really don't think it would be optimal for the docs to document anything less

NotNightSky:
> The recommended solution here is to use a 3rd party library. I really don't think it would be optimal for the docs to document anything less

~~At this point I don't think I even care because its_miroma, cassianc and celdaemon did not have any solutions, so whatever I see suggested next is going to be on the docs~~

I am going to write about world placement via 3rd party library.

NotNightSky:
@Wind292 could you please check the PR on the draft branch.

NotNightSky:
At first glance it seems like it is possible to add biomes to The Nether and The End according to the Biome API.
[NetherBiomes.java](https://github.com/FabricMC/fabric-api/blob/3e8afef37972285085fd0f7e414bcfcca487ed3c/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/NetherBiomes.java) and [EndBiomes.java](https://github.com/FabricMC/fabric-api/blob/3e8afef37972285085fd0f7e414bcfcca487ed3c/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/TheEndBiomes.java)

NotNightSky:
@Wind292 Hey, I just realized while experimenting with biome generation that we only have a guide for overworld biome and neither for end nor for nether. So, could you please look into that, if you have free time that is.

And let me know If you do start please.

Wind292:
> @Wind292 Hey, I just realized while experimenting with biome generation that we only have a guide for overworld biome and neither for end nor for nether. So, could you please look into that, if you have free time that is.
> 
> And let me know If you do start please.

@NotNightSky, I don't think there is a real difference between a "Nether biome" and an "Overworld biome". They are both just the same `Biome`and can be configured to spawn in whatever dimension you want.

NotNightSky:
> > @Wind292 Hey, I just realized while experimenting with biome generation that we only have a guide for overworld biome and neither for end nor for nether. So, could you please look into that, if you have free time that is.
> > And let me know If you do start please.
> 
> @NotNightSky, I don't think there is a real difference between a "Nether biome" and an "Overworld biome". They are both just the same `Biome`and can be configured to spawn in whatever dimension you want.

Surface rules are different and If not configured properly, the biome will spawn above the bedrock layer (that is what I found out during testing)

reference/latest/src/main/java/com/example/docs/ExampleMod.java
```diff
@@ -73,5 +81,35 @@ public void onInitialize() {
...
+										Climate.Parameter.span(-1.0f, -0.5f),
+										Climate.Parameter.span(-0.3f, 1.0f),
+										Climate.Parameter.span(-0.375f, 0.50f),
+										Climate.Parameter.point(0.0f),
+										Climate.Parameter.span(-0.5f, 1.0f),
+										0L));
+		// :::world-gen-biomes:overworld-addition
+
+		// :::world-gen-biomes:nether-addition
+		BiomePlacement.addNether(Biomes.ICE_SPIKES,
```
NotNightSky, line 105: The Nether biome can be added via `NetherBiomes.addNetherBiome` according to the [Fabric API](https://github.com/FabricMC/fabric-api/blob/26.2/fabric-biome-api-v1/src/main/java/net/fabricmc/fabric/api/biome/v1/NetherBiomes.java)
And I think we should use what is already available in the Fabric API.

Wind292: Yeah that should be mentioned. I'll add that right now

develop/data-generation/biome-generation.md
```diff
@@ -0,0 +1,76 @@
...
+To change what biomes can spawn in the overworld in minecraft you will need\* a external mod to manage the generation of biomes as the Fabric API does not implement it itself.
+
+::: info
+
+\* It is possible to add your own custom biomes directly to the overworld/end with mixins by overwriting the target dimension's data file, but it is not recommended. If you do so the compatibility with other mods' biomes will be terrible. Though adding to the Nether specifically [_is supported_](#nether-biome-fabricapi).
+
+:::
+
+The most popular libraries are [Biolith](https://github.com/TerraformersMC/Biolith) and [Terrablender](https://github.com/glitchfiend/terrablender).
+In this guide we are going to be using Biolith because it is lighter-weight and more simple to use than Terrablender.
```
bluebear94, line 27:
```suggestion
In this guide, we'll use Biolith because it is more lightweight and simpler to use than Terrablender.
```

develop/data-generation/dimension-creation.md
```diff
@@ -0,0 +1,589 @@
...
+The creation of Dimensions is broken down into 2 parts:
+
+- **DimensionType**: This defines the features of the dimension. i.e, hasFixedTime, hasSkyLight, minY, etc.
+- **Level Stem (Dimension)**: This defines the biomes, noise settings and dimensionType.
+
+::: info INFO
+
+A dimension may or may not have a custom biome. i.e, the user can choose to have singular or multiple vanilla biome.
```
bluebear94, line 31:
```suggestion
A dimension may or may not have custom biomes; it could contain one or more vanilla biomes.
```

develop/data-generation/biome-generation.md
Patbox: Might be also worth mentioning https://modrinth.com/mod/lithostitched as it's also quite popular

develop/data-generation/dimension-creation.md
```diff
@@ -0,0 +1,589 @@
...
+
+@[code lang=json](@/reference/latest/src/main/generated/data/example-mod/dimension_type/example_dimension_type.json)
+
+:::
+
+Now it is time to create the Level Stem.
+
+## Level Stem Provider {#levelstemprovider}
+
+::: warning WARNING
```
bluebear94, line 102: This message doesn’t seem to justify being a warning; I think it belongs as an info instead.

develop/data-generation/dimension-creation.md
```diff
@@ -0,0 +1,589 @@
...
+
+::: info INFO
+
+A dimension may or may not have a custom biome. i.e, the user can choose to have singular or multiple vanilla biome.
+
+:::
+
+## Dimension Type Provider {#dimensiontypeprovider}
+
+::: info INFO
```
bluebear94, line 37:
```suggestion
::: info
```

The `INFO` isn’t needed as the title will be “INFO” by default.

develop/data-generation/dimension-creation.md
```diff
@@ -0,0 +1,589 @@
...
+
+### Stem Customization {#stem-customization}
+
+The dimension stem can now be created.
+
+The code below will create a dimension with the same noise map as the overworld and have a single biome which we created as an example.
+
+@[code lang=java transcludeWith=:::datagen-dimension:dimensionStemCustomization](@/reference/latest/src/client/java/com/example/docs/datagen/ExampleModDimensionStemProvider.java)
+
+::: info TIP
```
bluebear94, line 126:
```suggestion
::: tip
```

There is a dedicated `tip` admonition.

develop/data-generation/dimension-creation.md
```diff
@@ -0,0 +1,589 @@
...
+
+- **biomes** :- The ID of the single biome in the world. In this guide, the ID is `examplemod:tater_biome` for which we use `TATER_BIOME_KEY`.
+
+- **lakes** :- Whether to generate lava lakes. If set to true, then lava lakes (lake_lava_underground and lake_lava_surface) generate even in biomes where lakes don't normally generate.
+
+- **features** :- Whether to generate biome-specific placed features.
+
+- **structure_override** :- List of structure sets to use. Can be an ID of structure set, a structure set tag, or a list of structure set IDs.
+
+Here are the following changes that need to be made inorder to use the flat generator.
```
bluebear94, line 208:
```suggestion
Here are the following changes that need to be made in order to use the flat generator.
```

develop/data-generation/dimension-creation.md
```diff
@@ -0,0 +1,589 @@
...
+
+```Text
+additional fields :
+├── biomes :- Any number of biome(s).
+└── scale :- (Optional, default is 2) Value between 0 and 62.
+```
+
+- **biomes** :- an ID, or a tag with #, or an array containing IDs.
+
+- **scale** :- Determines the size of the checkerboard grid. A scale of 0 means each cell of the grid is one chunk wide. is Value^2 chunks.
```
bluebear94, line 277:
```suggestion
- **biomes** :- an identifier of a biome, a tag prefixed with `#`, or an array of identifiers.
- **scale** :- Determines the size of the checkerboard grid. Each cell of the grid is `1 << scale` chunks wide.
```

develop/data-generation/dimension-creation.md
```diff
@@ -0,0 +1,589 @@
...
+
+Here are the following changes that need to be made inorder to use the checkerboard biome source.
```
bluebear94, line 279:
```suggestion
Here are the following changes that need to be made in order to use the checkerboard biome source.
```

develop/data-generation/dimension-creation.md
```diff
@@ -0,0 +1,589 @@
...
+
+    LevelStem exampleDimensionStem = new LevelStem(
+        dimensionTypes.getOrThrow(ExampleModDimensionTypeProvider.EXAMPLE_DIMENSION_TYPE_KEY),
+        chunkGenerator
+    );
+```
+
+The checkerboard biome source creates a pattern of biomes in a checkerboard layout.
+In the example above, we have used `TATER_BIOME_KEY` and `Biomes.PLAINS` as the biomes with a scale of 2.
+This means that the checkerboard pattern will have each biome take up a 4x4 chunk area.
```
bluebear94, line 303:
```suggestion
This means that the checkerboard pattern will have each biome take up a 4×4 chunk area.
```

its-miroma: hmm, I wonder where we should draw the line between typographically correct and standard keyboard characters.. For example, fancy quotes and ellipses are currently banned in markdownlint, and em-dashes have unfortunately become a shibboleth for the uninitiated. Afaik, the Docs should be entirely ASCII currently (except translations).

Should we use em- and en-dashes? Product symbols? Quotation marks? Section signs? There are a lot of symbols that we could use... 
https://developer.gnome.org/hig/guidelines/typography#take-advantage-of-unicode

develop/data-generation/dimension-creation.md
its-miroma: bitwise operators aren't obvious to every reader

bluebear94: Then what about:

```suggestion
- **biomes** :- an identifier of a biome, a tag prefixed with `#`, or an array of identifiers.
- **scale** :- Determines the size of the checkerboard grid. Each cell of the grid is `2 ** scale` chunks wide; that is, a `scale` of `0` gives 1-chunk wide cells, and increasing `scale` by 1 doubles the side length.
```

PR #557, opened 2026-04-03
Loader: Intro, `fabric.mod.json`, Dependency Overrides

cassiancc:
Updated and revised ports of the [Fabric Loader](https://wiki.fabricmc.net/documentation:fabric_loader) and [fabric.mod.json](https://wiki.fabricmc.net/documentation:fabric_mod_json), and [Dependency Overrides](https://wiki.fabricmc.net/tutorial:dependency_overrides) pages from the wiki. I've opted not to include the more technical aspects of the Fabric Loader article, as I do not consider them relevant to the Docs and cannot speak to their accuracy, but that can be added back later if desired.

PR #567, opened 2026-04-08
Class Tweakers: Enum Extensions

MildestToucan:
This PR aims to document the process of using Mixin and Class Tweaker's new feature: Enum Extension.

Before this PR is merged I would like to, on top of the usual reviewing process, get a greenlight from LlamaLad7 and EarthComputer to ensure accuracy. Additionally the question of whether to release the page before the feature is in a stable Loader version should be resolved.

This PR is at the time of writing going to be created as a Draft as I think more adjustments are needed, and pitfalls that can arise from extended enums (notably with `switch` expressions) still need to be added to the page.

Otherwise most of the knowledge about enum extension should be covered.


develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,90 @@
...
+If the enum entries need fields or to implement methods, you should shadow the relevant constructor by creating a matching one and annotating it with
+`@Shadow`. For example, adding a new `RecipeCategory` entry would look like this:
+
+<<< @/reference/latest/src/main/java/com/example/docs/mixin/class_tweakers/RecipeCategoryMixin.java#enum-extension-ctor-impls-example-mixin
+
+## Making the Class Tweaker Entry {#making-the-class-tweaker-entry}
+
+If you are targeting a Minecraft enum, you can use a class tweaker entry to make the added entry appear in the target enum.
+
+To opt into this feature, remember to use Loom 1.16 or above, and to have the file [header](../class-tweakers/index#file-format) have its
```
cassiancc: Have its...?

MildestToucan: Oh boy I really need to double check myself. Seems like I was a bit hasty with some phrasing lmao. I'll have that fixed soon.

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,90 @@
...
+title: Enum Extension
+description: Learn how to add entries to enums with Mixin and Class Tweakers.
+authors:
+  - MildestToucan
+---
+
+Enum extension is a Mixin feature that can reliably add new entries to an enum.
+
+When targeting Minecraft enums, the mixin can be combined with [class tweaking](../class-tweakers) to display the new enum entries
+to the decompiled source. This can be [transitive](../class-tweakers/index#transitive-entries) to allow mods depending on yours to see your added entries.
```
its-miroma:
```suggestion
When targeting Minecraft enums, you can use mixins together with [class tweaking](../class-tweakers) to append new enum entries
to the decompiled source. If that is set to be [transitive](../class-tweakers/index#transitive-entries), mods that depend on yours will also see your added entries.
```

alternatively, if "display" is the intended verb in this context:

```suggestion
When targeting Minecraft enums, you can use mixins together with [class tweaking](../class-tweakers) to display new enum entries
in the decompiled source. If that is set to be [transitive](../class-tweakers/index#transitive-entries), mods that depend on yours will also see your added entries.
```

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,90 @@
...
+
+::: warning
+
+Enum extension requires Loader 0.19.0 for Mixin support and Loom 1.16 for class tweaker support.
+
+Additionally, class tweaker file headers must specify a version of `v2` to use enum extensions.
```
its-miroma:
```suggestion
Additionally, class tweaker file headers must specify `v2` as the version to use enum extensions.
```

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,90 @@
...
+
+:::
+
+## Creating the Mixin {#creating-the-mixin}
+
+Before creating the mixin class, make sure that you are depending explicitly on Loader 0.19.0 or above
+in your `fabric.mod.json` file:
```
its-miroma:
```suggestion
Before creating the mixin class, make sure that Loader 0.19.0 or above is an explicit dependency
in your `fabric.mod.json` file:
```

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,91 @@
...
+
+```json:no-line-numbers
+...
+"depends": {
+  ...
+  "fabricloader": ">=0.19.0"
+  ...
+}
+...
```
its-miroma, line 38: did you know that the plurals of Ellipsis and of Ellipse are both Ellipses? This is a case of [polysemy](https://en.wikipedia.org/wiki/Polysemy), which is when a word has multiple different meanings that have a common origin. The common origin is the Greek [ἔλλειψις](https://en.wiktionary.org/wiki/%E1%BC%94%CE%BB%CE%BB%CE%B5%CE%B9%CF%88%CE%B9%CF%82), which is also the origin of the word Eclipse (the kind of eclipse that the Sun does every few years in bowing respect of the ominous IDE. Obviously, I'm referring to Android Studio, which when run on my laptop could easily replace the sun in providing heat for the Earth).

Polysemes are not to be confused with [Homonyms](https://en.wikipedia.org/wiki/Homonym), which happen coincidentally. An example of the latter is "left" (past of leave) and "left" (political orientation). Instead, polysemes can be traced to a common origin.

However, be careful! According to most sources I could find online, "Ellipses" is pronounced differently depending on the meaning:
- if it's plural for ellipsis (which is read eh-llip-sis): eh-llip-sees
- if it's plural for ellipse (which is read eh-llips): eh-llip-sis

That means, apparently, that ellipsis and ellipses (the shapes) are [homophones](https://en.wikipedia.org/wiki/Homophone), also known as heterographs. Although Wikipedia says homophones are a type of homonymy, which we said does not apply in this case, so idrk

this isn't really a review comment, i'm realizing just now. welp, I guess it's a freebie that you can mark as resolved!

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+Even if you are using the correct Loader version as a Gradle dependency, you must explicitly depend on it in
+order to opt into new Mixin features.
```
its-miroma:
```suggestion
order to opt into this Mixin feature.
```

shiny new isn't new forever: https://en.wikipedia.org/wiki/Pont_Neuf
> The **Pont Neuf** (French for "New Bridge") is the oldest standing bridge across the river Seine in Paris, France.

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,91 @@
...
+
+To make an enum extension, create a `@Mixin`-annotated enum class, and add your constants to it as if they were
+part of the targeted enum class. For example, adding a new `RecipeBookType` entry would look like this:
```
its-miroma:
```suggestion
To make an enum extension, create an `enum`, annotate it with `@Mixin`, and add your constants to it as if they were
part of the targeted enum class. For example, let's add a new entry to `RecipeBookType`:
```

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,91 @@
...
+
+Added enum constants should always be prefixed by your mod id to ensure uniqueness. To preserve capitalization, the prefix would typically be
+`MODID_`.
+
+:::
+
+If the enum entries need fields or to implement methods, you should shadow the relevant constructor by creating a matching one and annotating it with
+`@Shadow`. For example, adding a new `RecipeCategory` entry would look like this:
+
+<<< @/reference/latest/src/main/java/com/example/docs/mixin/class_tweakers/RecipeCategoryMixin.java#enum-extension-ctor-impls-example-mixin
```
its-miroma:
```suggestion
::: warning IMPORTANT

You should always prefix the enum constants you add with your mod ID to ensure uniqueness. For these Docs, we'll use `EXAMPLE_MOD_`.

:::

### Additional Fields and Methods {#additional-fields-and-methods}

If the targeted enum declares additional fields or method implementations, you'll have to shadow the relevant constructor.

For example, let's add a new `RecipeCategory` entry. Create a constructor matching the target, and annotate it with `@Shadow`.

<<< @/reference/latest/src/main/java/com/example/docs/mixin/class_tweakers/RecipeCategoryMixin.java#enum-extension-ctor-impls-example-mixin
```

imo it reads better like this, without fore@Shadowing (pun intended) the next section

develop/class-tweakers/index.md
```diff
@@ -25,9 +25,11 @@ Class tweaker files are conventionally named after your modid, `example-mod.clas
 The file must have the following header as its first line:
 
 ```classtweaker
-classTweaker v1 official
+classTweaker  v1  official
 ```
 
+Some features may require you to use a version number higher than `v1`, which will be specified in their dedicated pages.
```
its-miroma:
```suggestion
Some features may require a version higher than `v1` - that will be mentioned in their respective pages.
```

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Making the Class Tweaker Entry {#making-the-class-tweaker-entry}
+
+If you are targeting a Minecraft enum, you can use a class tweaker entry to make the added entry appear in the target enum.
```
its-miroma:
```suggestion
If you are targeting a Minecraft enum, you can use a class tweaker entry to actually modify the target enum.
```

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,91 @@
...
+
+To opt into this feature, remember to use Loom 1.16 or above, and to have the file [header](../class-tweakers/index#file-format) version as `v2`.
```
its-miroma:
```suggestion
To opt into this feature, remember to use Loom 1.16 or above, and to set the [file header version](../class-tweakers/index#file-format) to `v2`.
```

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Applying Changes {#applying-changes}
+
+To see your added enum entries in the decompiled source, you must refresh your Gradle project by [regenerating sources](../getting-started/generating-sources).
+If modifications do not appear, you can try [validating](../class-tweakers/index#validating-the-file) the file and checking if any errors appear.
+
+The added entries should now appear in the decompiled source. If the enum has fields or other elements for its entries to implement, those will not appear for your
+added entry, and are instead left to your mixin to implement at runtime.
+
+You can now use the enum constant from your code:
```
its-miroma:
```suggestion
You'll have to refresh your Gradle project by [regenerating sources](../getting-started/generating-sources) before you can see your added enum entries in the decompiled source.
If modifications do not appear, you can try [validating](../class-tweakers/index#validating-the-file) the file and checking if any errors appear.

::: info

You will not see [fields, methods or other elements](#additional-fields-and-methods) in the decompiled source code. That's because those are handled by the mixin, and are only applied at runtime.

:::

You can now use the enum constant in your code:
```

its-miroma (changes requested):
i'm sorry @cassiancc, I'll review #557 right now I promise

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,91 @@
...
+---
+
+Enum extension is a Mixin feature that can reliably add new entries to an enum.
+
+When targeting Minecraft enums, the mixin can be combined with [class tweaking](../class-tweakers) to display the new enum entries
+to the decompiled source. This can be [transitive](../class-tweakers/index#transitive-entries) to allow mods depending on yours to see your added entries.
+
+::: warning
+
+Enum extension requires Loader 0.19.0 for Mixin support and Loom 1.16 for class tweaker support.
```
cassiancc:
```suggestion
Enum extension requires at least Loader 0.19.0 for Mixin support and at least Loom 1.16 for class tweaker support.
```

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,91 @@
...
+...
+"depends": {
+  ...
+  "fabricloader": ">=0.19.0"
+  ...
+}
+...
+```
+
+Even if you are using the correct Loader version as a Gradle dependency, you must explicitly depend on it in
```
cassiancc:
```suggestion
Even if you are using the correct Loader version as a Gradle dependency, you must explicitly depend on at least version 0.19.0 in
```

develop/class-tweakers/enum-extension.md
cassiancc: Maybe "the features introduced in this version"?

cassiancc (changes requested):

develop/class-tweakers/enum-extension.md
its-miroma: This is the Enum Extension article specifically, that's why I restricted it to the singular

develop/class-tweakers/enum-extension.md
MildestToucan: My quota is gonna benefit so much from this. Thanks Miroma!

develop/class-tweakers/enum-extension.md
MildestToucan: Yeah Miroma's suggestion makes the most sense to me here. We could add something about the loader dep being related to Mixin feature opt-in on the FMJ page at some point, if that's not already planned.

develop/class-tweakers/enum-extension.md
MildestToucan: I'm not sure about this one. The enum *is* being "actually" modified by the Mixin at runtime. 

Class tweaking only takes care of making it appear in the decompiled source so it can be referenced directly and be seen by your dependents.

The suggested phrasing here imo is misleading as it implies that without class tweaking, you're not "actually" modifying the enum.

its-miroma: "to visibly modify the target enum [in the decompiled source]"?

MildestToucan: Yeah, "to visibly modify the target enum in the decompiled source" sounds good to me.

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,94 @@
...
+  "fabricloader": ">=0.19.0"
+  ...
+}
+...
+```
+
+Even if you are using the correct Loader version as a Gradle dependency, you must explicitly depend on at least version 0.19.0 in
+order to opt into this Mixin feature.
+
+To make an enum extension, create an `enum`, annotate it with `@Mixin`, and add your constants to it as if they were
```
cassiancc:
```suggestion
To make an enum extension, create an `enum` in your `mixin` package, annotate it with `@Mixin`, and add your constants to it as if they were
```

cassiancc (changes requested):
Overall, I think this is really good. I'll let the mixin experts bikeshed any technical details, but wording-wise I think this fits in the docs.

develop/class-tweakers/enum-extension.md
Jab125: imo "mixin" is better than "`mixin`"

MildestToucan: Will implement this without the monospacing as Jab125 is suggesting. Technically the mixin package could be named other things, so monospace might be a little misleading.

reference/latest/src/main/java/com/example/docs/enum_extension/ExampleModEnumExtension.java
```diff
@@ -0,0 +1,40 @@
...
+	// #endregion enum-extension-added-constant-no-ct-usage-example
+
+	// #region enum-extension-problematic-switch-expr-example
+	void exampleProblematicSwitch(RecipeBookType recipeBookType) {
+		String s = switch (recipeBookType) {
+			case SMOKER -> "smoker";
+			case CRAFTING -> "crafting";
+			case FURNACE -> "furnace";
+			case BLAST_FURNACE -> "blast_furnace";
+			case EXAMPLE_MOD_RECIPE_BOOK_TYPE -> "examplemod_recipe_book_type";
```
its-miroma:
```suggestion
			case EXAMPLE_MOD_RECIPE_BOOK_TYPE -> "example_mod_recipe_book_type";
```

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,127 @@
...
+<<< @/reference/latest/src/main/java/com/example/docs/enum_extension/ExampleModEnumExtension.java#enum-extension-added-constant-usage-example
+
+If you are only adding it with a mixin and it is not in the decompiled source, get it by calling `Enum#valueOf`:
+
+<<< @/reference/latest/src/main/java/com/example/docs/enum_extension/ExampleModEnumExtension.java#enum-extension-added-constant-no-ct-usage-example
+
+## Pitfalls {#pitfalls}
+
+Enum extension by itself makes no guarantees that your added entry will not break anything. You should find the modified enum's usages to try and
+prevent issues where possible. If an issue that would cause a crash arises and it cannot be fixed, it may be best to not use enum extension at all.
```
its-miroma:
```suggestion
Enum extension cannot guarantee that entries you add will not break anything.

It's your responsibility to go through the target enum's usages, and try to
prevent issues where possible. If you are unable to solve some, and crashes arise, it might be best to not use enum extension at all.
```

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,127 @@
...
+
+This section goes over patterns to look out for and to avoid in your own code, but it is in no way exhaustive.
```
its-miroma:
```suggestion
This section goes over some patterns to watch out for and avoid when extending enums, but it is not exhaustive.
```

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,127 @@
...
+
+You will not see [fields, methods or other elements](#additional-fields-and-methods) in the decompiled source code. That's because those are handled by the mixin, and are only applied at runtime.
+
+:::
+
+You can now use the enum constant in your code:
+
+<<< @/reference/latest/src/main/java/com/example/docs/enum_extension/ExampleModEnumExtension.java#enum-extension-added-constant-usage-example
+
+If you are only adding it with a mixin and it is not in the decompiled source, get it by calling `valueOf`:
```
its-miroma:
```suggestion
If you are only adding it with a mixin and it is not in the decompiled source, obtain it by calling `valueOf`:
```

Nit: "get it" is correct, but it can also mean "understand it" so I avoided it.

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,127 @@
...
+
+Enum extension by itself makes no guarantees that your added entry will not break anything. You should find the modified enum's usages to try and
+prevent issues where possible. If an issue that would cause a crash arises and it cannot be fixed, it may be best to not use enum extension at all.
+
+This section goes over patterns to look out for and to avoid in your own code, but it is in no way exhaustive.
+
+### Switch Expressions {#switch-expressions}
+
+Switches are often used to handle enum constants, this can lead to a crash if a switch expression has no `default` case to handle
+entries added by mods. For example, if we had the following switch expression:
```
its-miroma:
```suggestion
Switch statements are often used to handle enum constants. Because of this, a crash can happen if a switch expression does not handle
entries added by other mods. For example, say we have the following switch expression:
```

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,127 @@
...
+
+<<< @/reference/latest/src/main/java/com/example/docs/enum_extension/ExampleModEnumExtension.java#enum-extension-problematic-switch-expr-example
+
+Even though we handle all the values that are in the Vanilla enum and our own, this would throw for any other mod's added entry.
```
its-miroma:
```suggestion
Notice how there is no `default` clause. Even though we handled all the values in the Vanilla enum, and our own, this would throw if another mod adds a different entry.
```

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,127 @@
...
+
+If the expression is from a mod, you should consider raising an issue to the developers to find a more compatible option.
+If the expression is in a Vanilla method or you are unable to get the issue fixed by the mod developers, a mixin is likely needed.
+
+There is no one way for mixins to modify switch expressions to avoid throws, and the approach should be made on a case-by-case basis.
```
its-miroma: I flipped the order of the paragraphs:

```suggestion
How can you prevent this? There is no single universal way to avoid crashes - your approach should be adapted on a case-by-case basis. In general, though:

- If the `switch` expression is in a vanilla method, you can use a mixin to edit it
- If the `switch` expression comes from a mod, you should try getting in contact with the developers, to work together a compatible approach. Otherwise, you might have to create a mixin on the other mod.
```

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Serialized Enums {#serialized-enums}
+
+Certain enums may have their entries serialized automatically, such as the static `Variants` enum in the `Axolotl` class. Adding an entry to these enums
+would make your entry serialized under Minecraft's namespace by the codec, on some versions having it be serialized based on a numerical ID.
+
+It is best to avoid extending enums whose entries are serialized like this entirely, and if possible looking for an API.
```
its-miroma:
```suggestion
Certain enums' entries get serialized automatically. An example is the `Variants` enum in the `Axolotl` class.

Extending these enums
would serialize your custom entry under Minecraft's namespace, and on some versions that might happen based on a numerical ID.
This is not great because it can affect the indices of all other entries.

It is best to avoid extending enums entirely if their entries are serialized like this. Instead, you might want to look for an API, if available.
```

correct me

its-miroma (changes requested):

develop/class-tweakers/enum-extension.md
MildestToucan: You made a typo on `contact` on line 120, you typed `cnotact` instead.

its-miroma: You dare use my own spell against me, Potter..

(jokes aside, good catch!)

its-miroma (dismissed):

reference/latest/src/main/java/com/example/docs/enum_extension/ExampleModEnumExtension.java
```diff
@@ -0,0 +1,35 @@
...
+	void exampleRecipeBookTypeMethod(RecipeBookType recipeBookType) {
+		if (recipeBookType == RecipeBookType.EXAMPLE_MOD_RECIPE_BOOK_TYPE) {
+			/* ... */
+		}
+	}
+	// #endregion enum-extension-added-constant-usage-example
+
+	// #region enum-extension-added-constant-no-ct-usage-example
+	void exampleRecipeBookTypeMethodNoCT(RecipeBookType recipeBookType) {
+		if (recipeBookType == RecipeBookType.valueOf("EXAMPLE_MOD_RECIPE_BOOK_TYPE")) {
```
LlamaLad7: Wouldn't do this, either do .equals on the name or save the result of valueOf in some helper class

cassiancc (dismissed):

its-miroma (dismissed):

cassiancc (dismissed):

modmuss50 (dismissed):

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,139 @@
...
+
+::: warning IMPORTANT
+
+You should always prefix the enum constants you add with your mod ID to ensure uniqueness. For these Docs, we'll use `EXAMPLE_MOD_`.
+
+:::
+
+### Additional Fields and Methods {#additional-fields-and-methods}
+
+If the targeted enum declares additional fields or method implementations, you'll have to shadow the relevant constructor.
```
LlamaLad7: Not really true, it can have fields and methods but a default constructor. The key points are that if you want to use a non-default constructor you have to shadow it, and separately if the enum has abstract methods you must shadow and implement those in your added entries.

LlamaLad7 (changes requested):
Would be good to explain and show an example of extending abstract enums?

Jab125 (approved):

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,144 @@
...
+
+::: warning IMPORTANT
+
+You should always prefix the enum constants you add with your mod ID to ensure uniqueness. For these Docs, we'll use `EXAMPLE_MOD_`.
+
+:::
+
+### Specifying Fields {#specifying-fields}
+
+If the target enum has fields to initialize and no default, no-arg, constructor, you must shadow a target class constructor to
```
LlamaLad7: Nit: fields aren't relevant, only the presence of a no-arg constructor is

LlamaLad7 (approved):

its-miroma (approved):

its-miroma (approved):

its-miroma:
just *had* to approve twice

cassiancc (approved):

develop/class-tweakers/enum-extension.md
```diff
@@ -0,0 +1,144 @@
...
+Switch statements are often used to handle enum constants. Because of this, a crash can happen if a switch expression does not handle
+entries added by other mods. For example, say we have the following switch expression:
+
+<<< @/reference/latest/src/main/java/com/example/docs/enum_extension/ExampleModEnumExtension.java#enum-extension-problematic-switch-expr-example
+
+Notice how there is no `default` clause. Even though we handled all the values in the Vanilla enum, and our own, this would throw if another mod adds a different entry.
+
+How can you prevent this? There is no single universal way to avoid crashes - your approach should be adapted on a case-by-case basis. In general, though:
+
+- If the `switch` expression is in a vanilla method, you can use a mixin to edit it
```
p1k0chu, line 133: wouldn't it be cool if there was any elaboration on how to edit such a switch? i think just a hint in the right direction should be good enough, e.g. whats the best practice for the injector and target (at least for the situation in the code above)

LlamaLad7: No universal advice exists, at best people can be pointed to the discord / the general mixin injector pages

PR #575, opened 2026-04-19
chore: bump to 26.1.2

cassiancc:
Bumps the docs version to the latest (and hopefully last) hotfix for 26.1, 26.1.2.


its-miroma (approved):
I read a suggestion about changing the version picker to say "26.1.x"—I disagree, because that gives off the impression that either version is fine to use. Furthermore, we haven't done that ever previously, not for 1.21-1.21.1, not for 1.21.9, etc. Maybe we can open an issue for that, but I'll go ahead and merge this for now.

PR #578, opened 2026-04-23
fix: chronic papercuts

its-miroma:
as usual, please review the commits one-by-one


cassiancc (changes requested):
This currently causes redirects in the sidebar to no longer function e.g. Migrating Mappings.

its-miroma:
@cassiancc I think this is because PR previews don't build older versions

cassiancc:
That's what I thought originally, but I reproduced it locally as well.

its-miroma:
I also expect that, `transformHead` does not run in the Dev server.
Does the issue appear with `pnpm build && pnpm preview` also?

cassiancc:
Not sure how to test `build`, but `preview` has an even stranger result.
image

its-miroma:
`pnpm build` and `pnpm preview` are two separate commands, you should run the first then the second

its-miroma:
@cassiancc you're right, I'm able to reproduce the issue. the redirect won't happen unless you refresh. I asked for clarification at VitePress

modmuss50 (dismissed):

cassiancc (dismissed):

modmuss50 (approved):

Issue #307, opened 2025-03-16
Provide Kotlin examples alongside Java


Marobim815:
Hello Fabric Docs Team,

I noticed that the Fabric Docs provides code examples only in Java. However, many mod developers use Kotlin for its conciseness and better compatibility with modern Minecraft development.

Currently, Kotlin users have to manually convert Java examples to Kotlin, which can be inconvenient and time-consuming. Adding Kotlin examples alongside Java ones would make the documentation more accessible and beginner-friendly for Kotlin developers.

Would it be possible to include Kotlin examples in the Fabric Docs? I believe this would be a great help to the growing Kotlin modding community. If needed, I would be happy to contribute by converting Java examples to Kotlin.

Looking forward to your thoughts. Thanks for your hard work on the documentation!

cputnam-a11y:
> Hello Fabric Wiki Team,
> 
> I noticed that the Fabric Wiki provides code examples only in Java. However, many mod developers use Kotlin for its conciseness and better compatibility with modern Minecraft development.
> 
> Currently, Kotlin users have to manually convert Java examples to Kotlin, which can be inconvenient and time-consuming. Adding Kotlin examples alongside Java ones would make the documentation more accessible and beginner-friendly for Kotlin developers.
> 
> Would it be possible to include Kotlin examples in the Fabric Wiki? I believe this would be a great help to the growing Kotlin modding community. If needed, I would be happy to contribute by converting Java examples to Kotlin.
> 
> Looking forward to your thoughts. Thanks for your hard work on the documentation!

Note that this repository is for the official fabric documentation, which is a different team from the fabric wiki. For conversation on the topic of the wiki, you might visit the wiki channel in the discord server

IMB11:
Adding Kotlin examples would effectively double the documentation effort, and the contributor team is already working hard to maintain the current Java examples. Kotlin expertise isn't universal among all of the contributors, which would limit who could contribute effectively to a dual-language documentation set.

Also a core skill for Kotlin mod developers - maybe just in general Kotlin development too, is the ability to interpret and adapt Java code. The Java examples serve as a solid foundation, and there are tools and resources available elsewhere to assist with Java-to-Kotlin conversion.

PR #580, opened 2026-04-25
Kotlin docs

MaxURhino:
I saw, that the Fabric docs didn't have a Kotlin tab, so I created one.


dicedpixels:
I don't think we need Kotlin docs. If someone is using Kotlin, they should already be familiar with Java and know well enough to convert Java code.

its-miroma:
see #307 for why Kotlin is not documented rn

MaxURhino:
Ohh. So, how do I close this pull request?

its-miroma:
you can find the "Close pull request" button under the comment box

image

PR #590, opened 2026-04-30
Fix syncing in counter block

CelDaemon:
Adds a small section about using update packets, allowing other players to see the changes to the counter block without relogging.

Also fixes the overlay message to only send locally on the client.

### DIFF
--- a/develop/blocks/block-entities.md
+++ b/develop/blocks/block-entities.md
@@ -106,11 +106,9 @@ Now, when a player logs in or moves into a chunk where the block exists, they wi
 
 ## Syncing Data {#syncing-data}
 
-While new players loading in the block will see the correct count, other players watching the interaction will not see the count update; causing a desync.
+While new players loading in the block will see the correct count, the count will not update for other players watching the interaction. This phenomenon is called a desync, and it occurs when the server has updated its state, but the clients haven't.
 
-To solve this, we can use block entity update packets.
-
-Override the `getUpdatePacket` method, and return a packet containing the block's data from our `getUpdateTag`.
+To solve this, we can use block entity update packets. Override the `getUpdatePacket` method, and return a packet containing the block's data from our `getUpdateTag`.
 
 <<< @/reference/latest/src/main/java/com/example/docs/block/entity/custom/CounterBlockEntity.java#update-packet
 
--- a/develop/blocks/block-entities.md
+++ b/develop/blocks/block-entities.md
@@ -2,6 +2,7 @@
 title: Block Entities
 description: Learn how to create block entities for your custom blocks.
 authors:
+  - CelDaemon
   - natri0
 resources:
   https://docs.neoforged.net/docs/blockentities/: Block Entities - NeoForge Docs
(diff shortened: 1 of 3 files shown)


reference/latest/src/main/java/com/example/docs/block/entity/custom/CounterBlockEntity.java
```diff
@@ -46,6 +50,18 @@ public void incrementClicks() {
...
 
+	// #region broadcast-update
+	@Override
+	public void setChanged() {
+		super.setChanged();
+
+		if (level != null) {
+			BlockState state = getBlockState();
+			level.sendBlockUpdated(worldPosition, state, state, Block.UPDATE_ALL);
+		}
```
its-miroma:
```suggestion
		if (level == null) return;

		BlockState state = getBlockState();
		level.sendBlockUpdated(worldPosition, state, state, Block.UPDATE_ALL);
```

reference/latest/src/main/java/com/example/docs/block/custom/CounterBlock.java
```diff
@@ -44,9 +44,12 @@ protected InteractionResult useWithoutItem(BlockState state, Level level, BlockP
...
 
 		counterBlockEntity.incrementClicks();
-		player.sendOverlayMessage(Component.literal("You've clicked the block for the " + counterBlockEntity.getClicks() + "th time."));
 
-		return InteractionResult.SUCCESS;
+		if (level.isClientSide()) {
+			player.sendOverlayMessage(Component.literal("You've clicked the block for the " + counterBlockEntity.getClicks() + "th time."));
+		}
+
+		return InteractionResult.SUCCESS.withoutItem();
```
its-miroma: what is `withoutItem`?

develop/blocks/block-entities.md
```diff
@@ -103,6 +104,22 @@ To fix this, we override `getUpdateTag`:
 
 Now, when a player logs in or moves into a chunk where the block exists, they will see the correct counter value right away.
 
+## Syncing Data {#syncing-data}
+
+While new players loading in the block will see the updated count, other players watching the interaction will not see the updated count; causing a desync.
+
+To solve this, we can use block entity update packets.
+
+Override the `getUpdatePacket` method, and return a packet containing the block's data from our `getUpdateTag`.
```
its-miroma:
```suggestion
To solve this, we can use a block entity update packet. Override the `getUpdatePacket` method and return a packet, which will contain the data from `getUpdateTag`:
```

develop/blocks/block-entities.md
```diff
@@ -103,6 +104,22 @@ To fix this, we override `getUpdateTag`:
 
 Now, when a player logs in or moves into a chunk where the block exists, they will see the correct counter value right away.
 
+## Syncing Data {#syncing-data}
+
+While new players loading in the block will see the correct count, other players watching the interaction will not see the count update; causing a desync.
```
its-miroma:
```suggestion
While new players loading in the block will see the correct count, for other players who are watching the interaction the count won't update. This phenomenon is called a desync, and it occurs when the server has updated its state, but the clients haven't.
```

Maybe instead of this explanation we should link to Networking

its-miroma (changes requested):

reference/latest/src/main/java/com/example/docs/block/entity/custom/CounterBlockEntity.java
its-miroma: whoever taught github to render that diff like that, they should get their teaching license revoked

reference/latest/src/main/java/com/example/docs/block/custom/CounterBlock.java
CelDaemon: Whoops, unrelated. It marks that an interaction succeeded but did not use the active item.

cassiancc (approved):

its-miroma (approved):

PR #600, opened 2026-05-08
Implement codec docs in datagen

CelDaemon:
Moves inline snippets in the codecs docs to the reference mod, and uses datagen to test the codecs and generate output.


develop/codecs.md
```diff
@@ -28,34 +29,19 @@ Since a few vanilla classes already have codecs defined, we can use those as an
...
 Now, let's say we want to serialize a `BlockPos` to JSON and back. We can do this using the codec statically stored at `BlockPos.CODEC` with the `Codec#encodeStart` and `Codec#parse` methods, respectively.
 
-```java
-BlockPos pos = new BlockPos(1, 2, 3);
+::: code-group
 
-// Serialize the BlockPos to a JsonElement
-DataResult<JsonElement> result = BlockPos.CODEC.encodeStart(JsonOps.INSTANCE, pos);
-```
+<<< @/reference/latest/src/client/java/com/example/docs/datagen/CodecExampleProvider.java#encode-blockpos [Java]
```
cassiancc, line 34: Should this be a Codec/Input/Output tab block as well?

its-miroma: iirc the output tab in this case is an unhelpful `{}`

CelDaemon: I'm not entirely sure. The goal of the first snippet is showing the basic usage of codecs, I think it benefits from showing the full code example. I may move the conversion to JsonElement from the second snippet to this one though.

CelDaemon: > iirc the output tab in this case is an unhelpful {}

It shouldn't be? It's a BlockPos.

cassiancc: The output tab is fine, I'm just suggesting splitting the two lines in Java to Codec/Input like the rest of the tabs.

cassiancc (dismissed):

reference/latest/src/client/java/com/example/docs/datagen/ExampleModCodecExampleProvider.java
its-miroma, line 1: Why did you break the `ExampleMod*Provider` pattern?

I personally agree that prefixing with `ExampleMod` seems redundant, but I never insisted because I thought it was standard practice in Java..

I'm also noticing that `ExampleModEnchantmentGenerator` is the only one that isn't a `Provider` - why.

its-miroma (dismissed):

reference/latest/src/client/java/com/example/docs/datagen/ExampleModCodecExampleProvider.java
CelDaemon: While creating the file I was thinking that it wasn't really part of normal datagen, but is instead its own thing on top of datagen. But you're right, it should match the other classes anyway.

As for the enchantments one, I genuinely have no clue, I think it was already like that before I worked on it? Yarn called data providers "generators".

its-miroma (approved):

PR #614, opened 2026-05-21
Remove magic numbers from DirtChestMenu

CelDaemon:


reference/latest/src/main/java/com/example/docs/menu/custom/DirtChestMenu.java
```diff
@@ -27,22 +34,27 @@ public DirtChestMenu(final int containerId, final Inventory inventory, final Con
...
-			}
-		}
+		this.add3x3Slots(container, SLOTS_START_X, SLOTS_START_Y);
 
 		// Add the player inventory slots.
-		this.addStandardInventorySlots(inventory, 8, 84);
+		this.addStandardInventorySlots(inventory, INVENTORY_START_X, INVENTORY_START_Y);
+	}
+
+	private void add3x3Slots(Container container, int left, int top) {
```
its-miroma: Why are you passing `left` and `top` and not `rows`, `columns`, `size`?

CelDaemon: Honestly, fair. This signature matches vanilla's `DispenserMenu#add3x3GridSlots`.
I suppose it's to be able to render multiple grids with different containers, though vanilla never takes advantage of this. Will fix.

develop/blocks/container-menus.md
```diff
@@ -34,6 +34,11 @@ First, we want to create a block and block entity; read more in the [Block Conta
 
 @[code transcludeWith=:::be](@/reference/latest/src/main/java/com/example/docs/block/entity/custom/DirtChestBlockEntity.java)
 
+In addition to the normal block entity methods, we need to override the `stillValid` method. This method will be called every tick to check if the player should be forced out of the menu.
```
its-miroma, line 37:
```suggestion
In addition to the standard block entity methods, we need to override the `stillValid` method. This method will be called every tick to check if the player should be kicked out of the menu.
```

I would suggest a short clip to show a player getting kicked out of the menu (for example, water pushing it too far back), but I know how annoying it is to embed videos.

reference/latest/src/main/java/com/example/docs/menu/custom/DirtChestMenu.java
```diff
@@ -8,41 +8,59 @@
...
 
 // :::menu
 public class DirtChestMenu extends AbstractContainerMenu {
+	private static final int SLOTS_ROWS = 3;
+	private static final int SLOTS_COLUMNS = 3;
+	private static final int SLOTS_COUNT = SLOTS_ROWS * SLOTS_COLUMNS;
+
+	private static final int CONTAINER_START = 0;
+	private static final int CONTAINER_END = SLOTS_COUNT;
+	private static final int INVENTORY_START = CONTAINER_START;
```
its-miroma: what?

reference/latest/src/main/java/com/example/docs/menu/custom/DirtChestMenu.java
```diff
@@ -56,12 +74,17 @@ public ItemStack quickMoveStack(Player player, int slotIndex) {
 		ItemStack stack = slot.getItem();
 		ItemStack clicked = stack.copy();
 
-		if (slotIndex < this.container.getContainerSize()) {
-			if (!this.moveItemStackTo(stack, this.container.getContainerSize(), this.slots.size(), true)) {
+		if (slotIndex < CONTAINER_END) {
+			// If the clicked slot is in the container, try moving the item to the player inventory.
+			// When moving to the player inventory, starting backwards from the last slots to the first.
+			if (!this.moveItemStackTo(stack, INVENTORY_START, INVENTORY_END, /* backwards: */ true)) {
```
its-miroma, line 80: I'm conflicted about the inlay hint. On one hand, the `true` parameter is not very clear. On the other hand, I worry the explanation (line 79) makes it even less clear.

its-miroma (changes requested):

reference/latest/src/main/java/com/example/docs/menu/custom/DirtChestMenu.java
CelDaemon: My mistake, should be fixed now

develop/blocks/container-menus.md
```diff
@@ -34,6 +34,15 @@ First, we want to create a block and block entity; read more in the [Block Conta
...
+We will use the default implementation of this method from `ContainerHelper`, which checks if our block entity still exists and if the player is within the interaction range.
+
+<<< @/reference/latest/src/main/java/com/example/docs/block/entity/custom/DirtChestBlockEntity.java#container-still-valid
+
+After implementing our menu, it will close automatically when the player is pushed away.
+
+<VideoPlayer src="/assets/develop/blocks/menu_still_valid.webm"/>
```
its-miroma:
```suggestion
<VideoPlayer src="/assets/develop/blocks/menu_still_valid.webm">The container menu closes when the player goes out of range</VideoPlayer>
```

its-miroma (changes requested):

reference/latest/src/main/java/com/example/docs/menu/custom/DirtChestMenu.java
CelDaemon: I have reworded the explanation, I hope this is a little better?

cassiancc (approved):

its-miroma (approved):

PR #618, opened 2026-05-24
Added Repetitive model generation to datagen

JonyBoy19:
Whenever you create multiple variants of the same item or block you have to copy and paste the code for data-gen and then replace the name for them over and over again as in the case when you add multiple tool/armor materials to the game. Due to this I created documentation that shows users how to do this more quickly when you have multiple variants of the same items and blocks that you need to create.

(The purpose of adding this documentation as sub tabs instead of adding it to the existing tabs is that it simply does not fit in well with the already existing documentation)


develop/data-generation/repetitive-block-generation.md
```diff
@@ -0,0 +1,51 @@
+---
+title: Repetitive Block Generation
+description: Learn how to make repetitive block model generation easier
+authors:
+  - JonyBoy19
+---
+
+When adding several different varrients of the same block to your block model data generation it can get quite
+repetetive and make it difficult to find everything when you need to make changes.
```
CelDaemon: `repetitive`? :3

JonyBoy19: like when you add blue_block, red_block, and green_block. In most cases these blocks are exactly the same just with different colors and even if the block properties are different almost all the time the block models are the same. This also applies to items.

.vitepress/sidebars/develop.ts
```diff
@@ -269,10 +269,22 @@ export default [
               {
                 text: "develop.data_generation.block_models",
                 link: "/develop/data-generation/block-models",
+                items: [
+                  {
+                    text: "develop.data_generation.repetitive-block-generation",
```
Jimmy474: add display name for this side bar element to `sidebar_translations.json` using the `text` as key and display name as value

.vitepress/sidebars/develop.ts
```diff
@@ -269,10 +269,22 @@ export default [
...
+                    link: "/develop/data-generation/repetitive-block-generation",
+                  },
+                ],
               },
               {
                 text: "develop.data_generation.item_models",
                 link: "/develop/data-generation/item-models",
+                items: [
+                  {
+                    text: "develop.data_generation.repetitive-item-generation",
```
Jimmy474: add display name for this side bar element to `sidebar_translations.json` using the `text` as key and display name as value

CelDaemon:
I'm really not sure if this is something that fits the docs / something we should encourage.
In addition, this isn't really something specific to MC and anyone who knows java will probably know how to do this.

JonyBoy19:
> I'm really not sure if this is something that fits the docs / something we should encourage. In addition, this isn't really something specific to MC and anyone who knows java will probably know how to do this.

While this method isn't exactly specific to Minecraft and is more specific to Java for anyone who wants to make a bigger mod something like this ends up being highly useful or else you are stuck with a lot of clutter that makes it hard to find what you're looking for. Of course this still might not be something that will fit well into the documentation but even so if modders don't learn it here then they're going to have to learn it somewhere eventually.

develop/data-generation/repetitive-block-generation.md
its-miroma: she was pointing out a typo

.vitepress/sidebars/develop.ts
its-miroma: text should use `_`, not `-`: `develop.data_generation.repetitive_block_generation`

develop/data-generation/repetitive-block-generation.md
JonyBoy19: oh sorry, I'm bad with grammar

JonyBoy19:
Okay, I've finished fixing all the bugs and errors. If anyone wants to review this pull request and decide if it's something worth adding, it should be ready.

PR #631, opened 2026-06-05
Developer Guides: Statistics

### LINKED ISSUE #624: Developer Guides: Statistics
Prior Art: [Fabric Wiki Article](https://wiki.fabricmc.net/tutorial:stats)

Documents the creation of, writing to, and reading from, player statistics.

Tenneb22:
resolve #624


develop/statistics.md
```diff
@@ -0,0 +1,33 @@
...
+title: Statistics
+description: Learn how to create and use custom player statistics.
+authors:
+  - bemoty
+  - cassiancc
+  - jummit
+  - Tenneb22
+---
+
+Statistics track time spend doing something, distance traveled or jumped, and how often the player interacted with blocks or performed certain actions.
```
cassiancc:
```suggestion
Statistics track time spent performing actions in the world. Vanilla includes statistics for common actions like jumping, traveling in boats, as well as counting interactions with blocks or usage of items. It's also possible to add your own statistic to count custom interactions.
```

develop/statistics.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Creating a Statistic {#creating-a-statistic}
+
+To add a custom statistic, create an identifier which will be used to register and increase the stat:
```
cassiancc:
```suggestion
To add a custom statistic, create an `Identifier` which will be used to register and increase the stat:
```

reference/latest/src/main/java/com/example/docs/block/ModBlocks.java
```diff
@@ -164,6 +166,10 @@ public class ModBlocks {
 	);
 	// #endregion waxcap_tinting
 
+	public static final Block STATISTICS_BLOCK = register(
+					"statistics_block", StatisticsBlock::new, BlockBehaviour.Properties.of(), true
```
CelDaemon: Checkstyle won't complain, but please fix indenting

reference/latest/src/main/java/com/example/docs/stats/ModStats.java
```diff
@@ -0,0 +1,30 @@
...
+import net.minecraft.core.Registry;
+import net.minecraft.core.registries.BuiltInRegistries;
+import net.minecraft.resources.Identifier;
+import net.minecraft.stats.StatFormatter;
+import net.minecraft.stats.Stats;
+
+// #region register
+public class ModStats {
+	// #region stat
+	public static final Identifier INTERACT_WITH_STATISTICS_BLOCK = registerCustomStat("interact_with_statistics_block", StatFormatter.DEFAULT);
```
CelDaemon: Can we just call it `register`?

CelDaemon: Or, we can maybe document registry stats too?

reference/latest/src/main/resources/fabric.mod.json
```diff
@@ -29,7 +29,8 @@
       "com.example.docs.recipe.ExampleModRecipes",
       "com.example.docs.entity.ExampleModEntity",
       "com.example.docs.gamerule.ExampleModGameRules",
-      "com.example.docs.conditions.ExampleModResourceConditions"
+      "com.example.docs.conditions.ExampleModResourceConditions",
+      "com.example.docs.stats.ExampleModStats"
```
CelDaemon, line 33: I'd rather not introduce more initialisers.

develop/statistics.md
```diff
@@ -0,0 +1,33 @@
...
+
+<<< @/reference/latest/src/main/java/com/example/docs/stats/ModStats.java#stat
+
+### Registration {#registration}
+
+Then register the stat:
+
+<<< @/reference/latest/src/main/java/com/example/docs/stats/ModStats.java#register
+
+When adding the stat to the statistic screen via `Stats.CUSTOM.get()`, you can also specify the stat formatter. It determines how the number is shown in the stat list. You can use `DEFAULT`, `DIVIDE_BY_TEN`, `DISTANCE` or `TIME`.
```
cassiancc:
```suggestion
When adding the stat to the Statistics screen via `Stats.CUSTOM.get()`, you can also specify the stat formatter, which determines how the number is displayed in the stat list. Vanilla provides the following formatters:
- `DEFAULT`: Displays the number directly.
- `DIVIDE_BY_TEN`: Displays the number as a decimal, divided by ten.
-  `DISTANCE`: Displays the number as distance: Depending on the size of the number, this will be shown in centimeters, meters, or kilometers.
- `TIME`: Displays the number as time. Depending on the size of the number, this will be shown in seconds, minutes, hours, or days.
```

cassiancc (changes requested):

reference/latest/src/main/resources/fabric.mod.json
cassiancc: This is still the accepted way to introduce concepts that require static initialization - this can be refactored this at a later point if needed, but is fine for now.

Tenneb22: true, but where should it be initialized instead

develop/statistics.md
```diff
@@ -0,0 +1,38 @@
...
+Statistics track time spent performing actions in the world. Vanilla includes statistics for common actions like jumping, traveling in boats, as well as counting interactions with blocks or usage of items. It's also possible to add your own statistic to count custom interactions.
+
+## Creating a Statistic {#creating-a-statistic}
+
+To add a custom statistic, create an `Identifier` which will be used to register and increase the stat:
+
+<<< @/reference/latest/src/main/java/com/example/docs/stats/ModStats.java#stat
+
+### Registration {#registration}
+
```
its-miroma:
```suggestion
```

develop/statistics.md
```diff
@@ -0,0 +1,38 @@
...
+- `DEFAULT`: Displays the number directly.
+- `DIVIDE_BY_TEN`: Displays the number as a decimal, divided by ten.
+- `DISTANCE`: Displays the number as distance: Depending on the size of the number, this will be shown in centimeters, meters, or kilometers.
+- `TIME`: Displays the number as time. Depending on the size of the number, this will be shown in seconds, minutes, hours, or days.
+
+## Using The Statistic {#using-the-statistic}
+
+To increment the statistic, for example when a player interacts with a block, you can use `Player#awardStat`:
+
+<<< @/reference/latest/src/main/java/com/example/docs/block/custom/StatisticsBlock.java#interact
```
its-miroma: I think this statistic (uses of a block) is already tracked by vanilla? I'm not sure.

I was thinking of a way to introduce a new stat: maybe "tallest tower", and it tracks how many consecutive Stat blocks are stacked on top of each other? Is that doable?

Maybe: when a Stat block is placed, it first checks if the player has already been awarded the height limit as the Tallest Tower. If not, it should recursively check the top and bottom blocks until it finds a non-Stat block, and sums those up?

its-miroma (changes requested):

develop/statistics.md
CelDaemon: Unfortunately stats don't really provide a way to save specific amounts. You can *add* any amount, but you're not really supposed to retrieve the stats like that.

its-miroma: I see... what about "Friendships made"? When you place a stat block, any neighboring stat blocks are its friends, so if there are four and you place one in the center, +5 is added to the friendship stat?

I realize this is overcomplicated, if you can make up a simpler example you're welcome to, just something distinct from what vanilla already has.

CelDaemon: Sounds doable

reference/latest/src/main/java/com/example/docs/block/custom/StatisticsBlock.java
```diff
@@ -0,0 +1,41 @@
...
+	public StatisticsBlock(Properties properties) {
+		super(properties);
+	}
+
+	// #region statistics_block
+	@Override
+	public void setPlacedBy(Level level, BlockPos pos, BlockState state, @Nullable LivingEntity by, ItemStack itemStack) {
+		super.setPlacedBy(level, pos, state, by, itemStack);
+
+		if (by instanceof Player player) {
```
its-miroma: I didn't know about this syntax to assign `player`! Does an early return work with it, or is it local to the `if` block only? For example:

```java
if (!(by instanceof Player player)) return;

// ...

player.awardStat(ModStats.FRIENDSHIPS_MADE, neighborCount);
```

its-miroma (dismissed):

reference/latest/src/main/java/com/example/docs/block/custom/StatisticsBlock.java
Tenneb22: This existed for a long time, I think, but the early return also works! Didn't know that either.

An early return would make it cleaner, so I'll change it..?

cputnam-a11y: Looks like the initial features came around java 14, and some more was added in java 21

CelDaemon: And mojang recently started using it :3

cassiancc (dismissed):

reference/latest/src/main/java/com/example/docs/stats/ModStats.java
```diff
@@ -0,0 +1,30 @@
...
+
+	private static Identifier register(String name, StatFormatter formatter) {
+		Identifier id = Identifier.fromNamespaceAndPath(ExampleMod.MOD_ID, name);
+		Registry.register(BuiltInRegistries.CUSTOM_STAT, name, id);
+		Stats.CUSTOM.get(id, formatter);
+		return id;
+	}
+	// #endregion register
+
+	public static void initialize() {
```
CelDaemon, line 26: I think we should specifically mention calling this, people forget to do that way too often.

Tenneb22: How about "Don't forget to initialize the ModStats class in your mod initializer!", would that be enough?

its-miroma: Yes, with a link to project-structure. Grep mod initializer to find how we've been doing it

develop/statistics.md
```diff
@@ -0,0 +1,44 @@
...
+
+Don't forget to initialize the `ModStats` class in your [mod's initializer](./getting-started/project-structure#entrypoints):
+
+<<< @/reference/latest/src/main/java/com/example/docs/stats/ModStats.java#initialize
+
+<<< @/reference/latest/src/main/java/com/example/docs/stats/ExampleModStats.java#initialize
+
+## Using the Statistic {#using-the-statistic}
+
+In this example, we will increase the statistic whenever the friends block is placed. We use the `Player#awardStat(stat, amount)` method to increment it by the amount of neighbors the block has when placed.
```
its-miroma:
```suggestion
For this example, we will create a Friends block, which makes friends with its neighbors. We shall track how many friendships the player has formed with the block.

To do this, we will use the `Player#awardStat(stat, amount)` method to increment the stat by the amount of neighbors the block has when placed:
```

develop/statistics.md
```diff
@@ -0,0 +1,44 @@
...
+- `DEFAULT`: Displays the number directly.
+- `DIVIDE_BY_TEN`: Displays the number as a decimal, divided by ten.
+- `DISTANCE`: Displays the number as distance: Depending on the size of the number, this will be shown in centimeters, meters, or kilometers.
+- `TIME`: Displays the number as time. Depending on the size of the number, this will be shown in seconds, minutes, hours, or days.
+
+Don't forget to initialize the `ModStats` class in your [mod's initializer](./getting-started/project-structure#entrypoints):
+
+<<< @/reference/latest/src/main/java/com/example/docs/stats/ModStats.java#initialize
+
+<<< @/reference/latest/src/main/java/com/example/docs/stats/ExampleModStats.java#initialize
```
its-miroma, line 36: Do we usually include both of these?

develop/statistics.md
```diff
@@ -0,0 +1,44 @@
...
+authors:
+  - bemoty
+  - cassiancc
+  - CelDaemon
+  - its-miroma
+  - jummit
+  - Tenneb22
+---
+
+Statistics track time spent performing actions in the world. Vanilla includes statistics for common actions like jumping, traveling in boats, as well as counting interactions with blocks or usage of items. It's also possible to add your own statistic to count custom interactions.
```
its-miroma:
```suggestion
Statistics (or Stats) are a way to keep track of certain actions or time that the player spends in the world. Vanilla tracks statistics for common actions like jumping, traveling in boats, interactions with blocks, usage of items and much more. It's also possible to add your own statistic to track some custom interaction.
```

develop/statistics.md
```diff
@@ -0,0 +1,44 @@
...
+
+<<< @/reference/latest/src/main/java/com/example/docs/stats/ExampleModStats.java#initialize
+
+## Using the Statistic {#using-the-statistic}
+
+In this example, we will increase the statistic whenever the friends block is placed. We use the `Player#awardStat(stat, amount)` method to increment it by the amount of neighbors the block has when placed.
+
+<<< @/reference/latest/src/main/java/com/example/docs/block/custom/FriendsBlock.java#friends_block
+
+You can also use `Player#awardStat(stat)` or `Player#resetStat(stat)` to increase the stat by 1 or reset it to 0, respectively.
```
its-miroma: Should we make it so that breaking a Friends block causes a civil war, and all friendships are immediately broken (by calling resetStat)?

its-miroma (changes requested):

develop/statistics.md
CelDaemon: We do, but usually the snippets don't intersect.

develop/statistics.md
```diff
@@ -0,0 +1,50 @@
...
+
+For this example, we will create a Friends block, which makes friends with its neighbors. We shall track how many friendships the player has formed with the block.
+
+To do this, we will use the `Player#awardStat(stat, amount)` method to increment the stat by the amount of neighbors the block has when placed:
+
+<<< @/reference/latest/src/main/java/com/example/docs/block/custom/FriendsBlock.java#friends_block
+
+You can also use `Player#awardStat(stat)` to increment the stat by 1.
+
+Since Friends block are very attached to each other, breaking one means breaking all friendships. Let's make it so that breaking a Friends block will reset the player's stat back to 0 by using `Player#resetStat()`:
```
its-miroma:
```suggestion
Since Friends blocks are very attached to each other, breaking one means breaking all friendships. Let's make it so that breaking a Friends block will reset the player's stat back to 0, by using `Player#resetStat()`:
```

i made a mistake

its-miroma (changes requested):

its-miroma (approved):

cassiancc (approved):

PR #636, opened 2026-06-07
Move code examples from Entity Attributes page into a Game Test

### LINKED ISSUE #535: Entity Attributes: Move Code to ExampleMod
Now that #267 is merged :tada:, including some Entity-related code, the code blocks we had embedded in the Entity Attributes file should be tracked in the reference mod, so they don't break with updates.

https://github.com/FabricMC/fabric-docs/blob/1dc533219d7715fe1fd27d76fed0ecf2175c3af5/develop/entities/attributes.md?plain=1#L61

Tenneb22:
fix #535

This is not the best implementation, but it's better than keeping them in the markdown files.
And they should be easy to move out of the ReferenceMethods class when needed.


reference/latest/src/main/java/com/example/docs/ReferenceMethods.java
```diff
@@ -70,4 +77,33 @@ public static void savedDataExampleScenario() {
 		});
 		// #endregion saved_data_example_scenario
 	}
+
+	// #region applying_entity_attributes
```
CelDaemon: This is dead code, can we show this using our custom entity?

CelDaemon (changes requested):

reference/latest/src/main/java/com/example/docs/ReferenceMethods.java
Tenneb22: Sure, but it doesn't use the custom attribute that was created in the guide. How about adding it to the golem, or is that unnecessary?

CelDaemon: That is what I'd suggest yeah.

reference/latest/src/main/java/com/example/docs/entity/MiniGolemEntity.java
```diff
@@ -45,12 +47,15 @@ public MiniGolemEntity(EntityType<? extends MiniGolemEntity> entityType, Level w
...
 	public static AttributeSupplier.Builder createCubeAttributes() {
 		return PathfinderMob.createMobAttributes()
 				.add(Attributes.MAX_HEALTH, 5)
 				.add(Attributes.TEMPT_RANGE, 10)
-				.add(Attributes.MOVEMENT_SPEED, 0.3);
+				.add(Attributes.MOVEMENT_SPEED, 0.3)
+				.add(ModAttributes.AGGRO_RANGE, 8.0);
 	}
+	// #endregion attributes
 	// #endregion registerclass
```
its-miroma, line 61: Is #registerclass impacted by this meaningfully?

its-miroma (approved):
This looks reasonable.

- Are we sure we have to place those last two methods in `ReferenceMethods`?
- ~~I'd rename that class to `Unreferenced`.~~
- ~~Can you confirm that `ReferenceMethods` is getting built? Are we sure that the tooling does not ignore unreferenced classes?~~

reference/latest/src/main/java/com/example/docs/entity/MiniGolemEntity.java
cassiancc: If people are copy/pasting from the first entity article, it may be unclear where `ModAttributes` comes from.

its-miroma: then I'd move the `AGGRO_RANGE` as the first attribute and exclude it from #registerclass.

reference/latest/src/main/java/com/example/docs/entity/MiniGolemEntity.java
```diff
@@ -45,12 +47,15 @@ public MiniGolemEntity(EntityType<? extends MiniGolemEntity> entityType, Level w
...
 	}
 
+	// #region attributes
 	public static AttributeSupplier.Builder createCubeAttributes() {
 		return PathfinderMob.createMobAttributes()
 				.add(Attributes.MAX_HEALTH, 5)
 				.add(Attributes.TEMPT_RANGE, 10)
-				.add(Attributes.MOVEMENT_SPEED, 0.3);
+				.add(Attributes.MOVEMENT_SPEED, 0.3)
+				.add(ModAttributes.AGGRO_RANGE, 8.0);
```
its-miroma:
```suggestion
				// #endregion registerclass
				.add(ModAttributes.AGGRO_RANGE, 8.0)
				// #region registerclass
				.add(Attributes.MAX_HEALTH, 5)
				.add(Attributes.TEMPT_RANGE, 10)
				.add(Attributes.MOVEMENT_SPEED, 0.3);
```

its-miroma (changes requested):

reference/latest/src/main/java/com/example/docs/entity/MiniGolemEntity.java
CelDaemon: Would be nice if we could also find a use for the other snippets though?
Maybe even a custom attribute.

Tenneb22:
> Can you confirm that `ReferenceMethods` is getting built? Are we sure that the tooling does not ignore unreferenced classes?

Yes, it's getting built, and checked by the tooling.

its-miroma (approved):

cassiancc (approved):

PR #642, opened 2026-06-14
Add more details to Custom Tools section

bluebear94:
Adds details about how to create the incorrect blocks tag.

Inspired by [NeoForge’s docs on the same topic](https://docs.neoforged.net/docs/items/tools#tags), which [ChampionAsh5357 graciously answered my inquiry about](https://github.com/neoforged/Documentation/issues/359#issuecomment-4699194102).

Also mention the special tool item classes `AxeItem`, `HoeItem`, and `ShovelItem` (NeoForge has `ItemAbility` to make this unnecessary, but Fabric doesn’t).

Fixes #622, supersedes and closes #623

PR #645, opened 2026-06-17
Port to 26.2

cassiancc:
TODO:
- [x] Document `ItemIds`, `BlockIds`, `BlockItemIds`, `FluidIds` migration
- [x] Document Criteria registration
- [x] Migrate block ids to `BlockIds`
- [x] Fix custom render pipeline (thanks @CelDaemon and @Jimmy474!)
- [x] Re-base on 26.1.
- [x] Fix any build errors/checkstyle (blocked, build is erroring due to framework issues)

PR #649, opened 2026-06-19
Revise custom tools section

bluebear94:
Fix things that weren’t caught in the previous PR, #642.


its-miroma (approved):

its-miroma:
Consider opening a similar PR to cassian's #645, targeting both latest and 26.1.2

PR #666, opened 2026-07-04
Note that mcsrc.dev supports all versions

cassiancc:
`mcsrc.dev` was recently updated to support unobfuscated versions, making this note unnecessary.

### DIFF
--- a/develop/class-tweakers/access-widening.md
+++ b/develop/class-tweakers/access-widening.md
@@ -121,7 +121,6 @@ Manually writing access widener entries is time-consuming and prone to human err
 
 ### mcsrc.dev {#mcsrc-dev}
 
-Available for all versions with an [unobfuscated JAR](../migrating-mappings/index#whats-going-on-with-mappings), namely 1.21.11 and above,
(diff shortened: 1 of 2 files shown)


Earthcomputer:
`mcsrc.dev` still does not support *all* Minecraft versions, it supports all unobfuscated versions and versions with Mojmap. These are:

- All versions c0.0.13a and below
- 1.14.4
- All versions 19w36a (a snapshot for 1.15) and above.

It still excludes:
- All versions from (not including) c0.0.13a up to (not including) 1.14.4
- 19w34a
- 19w35a

cassiancc:
It supports all versions that Docs do, so I don't think specific clarification is necessary.

MildestToucan (approved):
Yeah, note no longer applies to mcsrc. 

It could be neat to also apply this change to the 1.21.11 page since it's quite small, but it's my understanding that docs changes aren't necessarily consistently backported unless it's particularly important, so I'll leave that matter up to the docs team.

cassiancc:
It's a trivial change to make, I'll just do it next time I'm at my laptop. Backporting docs right now just sucks UX-wise.

its-miroma (approved):

PR #674, opened 2026-07-12
feat(ides): add zed

itzzjustmateo:


MildestToucan:
I'm not sure this is worthwhile. I'm generally not against covering more stuff in docs, but I don't think I've seen much of anyone using Zed for mod development.

I'm not entirely against adding this to the docs, but it does seem unnecessary for little benefit, and if Zed has notable changes in its functionality over time (it's still a relatively young editor, so it wouldnt be surprising) practically no maintainer would be able to update the page because of how rare Zed is as an editor in the community. 

I encourage others to weigh in on this, but I don't feel like this would be a worthwhile thing to add to the already high and increasing count of pages we have to maintain over time.

itzzjustmateo:
> I'm not sure this is worthwhile. I'm generally not against covering more stuff in docs, but I don't think I've seen much of anyone using Zed for mod development.
> 
> I'm not entirely against adding this to the docs, but it does seem unnecessary for little benefit, and if Zed has notable changes in its functionality over time (it's still a relatively young editor, so it wouldnt be surprising) practically no maintainer would be able to update the page because of how rare Zed is as an editor in the community.
> 
> I encourage others to weigh in on this, but I don't feel like this would be a worthwhile thing to add to the already high and increasing count of pages we have to maintain over time.

Makes sense, I just thought I'd open a PR for this, as I use Zed alot for everything, but I mean there is also no page for Vim and NeoVim for a reason, nobody uses it for Mod-Development :rofl: I understand you thought, it is not a problem if this gets closed

PR #682, opened 2026-07-22
Add `networkSynchronized` to data component docs

ekulxam:
And a few misc fixes

Will be backported later by me (theoretically in this PR), I just want some preliminary reviews to ensure my writing and logic are sound.

### DIFF
--- a/develop/items/custom-data-components.md
+++ b/develop/items/custom-data-components.md
@@ -46,11 +46,19 @@ As an example, let's create an `Integer` value that will track how many times th
 
 <<< @/reference/latest/src/main/java/com/example/docs/component/ModComponents.java#integer_component
 
-You can see that we're now passing `<Integer>` as our generic type, indicating that this component will be stored as a single `int` value. For our codec, we are using the provided `ExtraCodecs.NON_NEGATIVE_INT` codec. We can get away with using basic codecs for simple components like this, but more complex scenarios might require a custom codec (this will be covered briefly later on). For our `StreamCodec` (sending the information to the client), we're using `ByteBufCodecs.VAR_INT`. A VarInt is a method of storing integers using a dynamic number of bytes. A smaller integer will be encoded in less bytes, while a larger integer will require more bytes.
+You can see that we're now passing `<Integer>` as our generic type, indicating that this component will be stored as a single `int` value.
+
+For our codec, we are using the provided `ExtraCodecs.NON_NEGATIVE_INT` codec. We can get away with using basic codecs for simple components like this, but more complex scenarios might require a custom codec (this will be covered briefly later on).
+
+For our `StreamCodec` (sending the information to the client), we're using `ByteBufCodecs.VAR_INT`. A VarInt is a method of storing integers using a dynamic number of bytes. A smaller integer will be encoded in less bytes, while a larger integer will require more bytes.
 
 ::: warning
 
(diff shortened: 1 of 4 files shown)


develop/items/custom-data-components.md
```diff
@@ -46,7 +46,13 @@ As an example, let's create an `Integer` value that will track how many times th
 
 <<< @/reference/latest/src/main/java/com/example/docs/component/ModComponents.java#integer_component
 
-You can see that we're now passing `<Integer>` as our generic type, indicating that this component will be stored as a single `int` value. For our codec, we are using the provided `ExtraCodecs.POSITIVE_INT` codec. We can get away with using basic codecs for simple components like this, but more complex scenarios might require a custom codec (this will be covered briefly later on).
+You can see that we're now passing `<Integer>` as our generic type, indicating that this component will be stored as a single `int` value. For our codec, we are using the provided `ExtraCodecs.NON_NEGATIVE_INT` codec. We can get away with using basic codecs for simple components like this, but more complex scenarios might require a custom codec (this will be covered briefly later on). For our `StreamCodec` (sending the information to the client), we're using `ByteBufCodecs.VAR_INT`. A VarInt is a method of storing integers using a dynamic number of bytes. A smaller integer will be encoded in less bytes, while a larger integer will require more bytes.
```
cassiancc: Nit: Why change to `ExtraCodecs.NON_NEGATIVE_INT`?

ekulxam: The documentation previously referred to `ExtraCodecs.POSITIVE_INT` but we were actually using `Codec.INT`, so somewhere along the line there was a desync.

`NON_NEGATIVE_INT` is more correct because it allows 0, and we explicitly set it to 0 in the code.

cassiancc (dismissed):

develop/items/custom-data-components.md
its-miroma:
```suggestion
You can see that we're now passing `<Integer>` as our generic type, indicating that this component will be stored as a single `int` value.

For our codec, we are using the provided `ExtraCodecs.NON_NEGATIVE_INT` codec. We can get away with using basic codecs for simple components like this, but more complex scenarios might require a custom codec (this will be covered briefly later on).

For our `StreamCodec` (sending the information to the client), we're using `ByteBufCodecs.VAR_INT`. A VarInt is a method of storing integers using a dynamic number of bytes. A smaller integer will be encoded in less bytes, while a larger integer will require more bytes.
```

develop/items/custom-data-components.md
```diff
@@ -46,7 +46,13 @@ As an example, let's create an `Integer` value that will track how many times th
...
+
+::: warning
+
+Do not think you can skip `networkSynchronized`! If you do not specify a `StreamCodec` in this manner, Minecraft will instead construct a more inefficient one and sync your component anyway. To prevent your component from being synced, you will need mixins, which is outside the scope of this tutorial.
```
its-miroma:
```suggestion
Do not think you can skip `networkSynchronized`!

If you do not specify a `StreamCodec` in this manner, Minecraft will instead construct a more inefficient one and sync your component anyway.

To prevent your component from being synced, you will need mixins, which is outside the scope of this tutorial.
```

its-miroma (changes requested):
lgtm

ekulxam:
Thanks for the reviews, time to backport

its-miroma (approved):

cassiancc (approved):

PR #687, opened 2026-07-25
Remove use of GLFW in KeyMapping page

cassiancc:
26.3 has removed GLFW, which means any mods using it for key mappings will have to switch over to Mojang's own `InputConstants`. As these input constants have existed at least as far back as the `KeyMapping` page, I see no reason why we can't recommend making use of them now, saving time porting.

### DIFF
--- a/develop/key-mappings.md
+++ b/develop/key-mappings.md
@@ -32,11 +32,11 @@ our key mapping at the same time.
 
 ::: info
 
-Note that the names of the key tokens (`GLFW.GLFW_KEY_*`) assume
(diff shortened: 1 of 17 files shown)


its-miroma (dismissed):
I assume that the mention of GLFW on the page will remain tho

dicedpixels (dismissed):

cassiancc:
Hm, that mention refers directly to the names of the tokens as if you're seeing them in the text block above. I think it might be clearer to change that.

its-miroma (approved):

PR #680, opened 2026-07-20
[EARLY DRAFT] New: Mixin Introduction and `mixins.json` config files

MildestToucan:
## Draft Notice

This is a ***VERY EARLY*** look at the upcoming Mixin introduction + Mixin config PR. It is by no means 
ready for review and contains little content.

This draft exists so early in the progression of this PR mainly to allow interested and eager parties to make suggestions and influence the shape of the pages as they are being made.

This is a relatively very large PR in terms of how much knowledge it will contain and how many structural decisions it will need to make for the future of Mixin on the Docs, so getting input and feedback as early as possible and throughout the process is critical.

Additionally, due to the scope and the importance of making Mixin docs right, this PR will also serve as a way for others interested in making a similar PR to concentrate their efforts and discussions on feedback, reviews, suggestions and decisions implemented by this PR, rather than dealing with the previous stale PRs, or making their own PRs and causing unorganized and duplicate work.

TL;DR: This is an extremely early draft, and I opened the draft PR that early to prevent future duplicate PR efforts on the topic, and to allow interested parties to weigh in on the discussion early on and influence the PR as it progresses.

## Draft Checklist

The PR aims to cover the following topics before being readied for review and promoted out of drafting:

- [ ] Introduction to Mixin as a concept: A high level overview of what Mixin iswhat it does and a (not over)simplified idea of how it works. This should defer back to the bytecode page (and if necessary amend it) whenever bytecode concepts are brought up to avoid redundancy.
- [ ] Mixin classes and injection basics: A beginner-friendly introduction to creating a very basic mixin class and injector, allowing a reader to learn basic syntax and mixin registration in preparation for further learning.
- [ ] A page detailing the different options of a `mixins.json` config file that are relevant to mod developers. This should defer to (and amend if necessary) the `fabric.mod.json` article for registering the config file itself for Fabric.

## Goals and Scope

This PR will **not** include "knowledge base" pages, meaning pages which document chunks of information in isolation from the introductory process. For example, this PR shouldn't include a dedicated `@Inject` page detailing the uses, attributes and precise effects of the injector.

Instead, this PR aims to focus on the resources and introductions necessary for a beginner to efficiently get help in dedicated support channels, be prepared for more advanced and specific reading added in future PRs, and understand the fundamental concepts around mixins.

In short, this PR aims to bring the necessary introductions and resources for beginners to more easily get past the initially very steep learning difficulties of Mixin.

## Related issues and PRs

Partially resolves #36, and by extension also relates to its parent issue #6 

[...]

PR #692, opened 2026-08-03
Document mixin compat level handling

dexman545:
Add documentation for mixin compat level, which is based on the mod's floader version dependency. The bytecode section seemed to be the most relevant existing page for this, and I don't think the list is long enough to warrant its own page yet.

### DIFF
--- a/develop/loader/fabric-mod-json.md
+++ b/develop/loader/fabric-mod-json.md
@@ -227,9 +227,8 @@ The following keys will accept a dictionary of dependencies. For more details on
 Fabric Loader will examine a mod's dependency on Loader to control changes to mixin behavior that
 may be incompatible with earlier or newer versions of Loader.
 
-The mods declared `depends` and `breaks` clauses in the [fabric.mod.json](../loader/fabric-mod-json#dependency-resolution) targeting `fabricloader`
-or `fabric-loader` (the former is preferred). The minimum compatible Loader version will determine
-the compatibility level.
+The mods declared `depends` and `breaks` clauses in the [fabric.mod.json](../loader/fabric-mod-json#dependency-resolution) targeting `fabricloader`. 
+The minimum compatible Loader version will determine the compatibility level.
 
 If no Loader dependency is specified, the minimum compatibility level (equivalent to Fabric Loader 0.9.2) is used.
 
--- a/develop/loader/fabric-mod-json.md
+++ b/develop/loader/fabric-mod-json.md
@@ -227,7 +227,7 @@ The following keys will accept a dictionary of dependencies. For more details on
 Fabric Loader will examine a mod's dependency on Loader to control changes to mixin behavior that
 may be incompatible with earlier or newer versions of Loader.
 
-The mods declared `depends` and `breaks` clauses in the [fabric.mod.json](../loader/fabric-mod-json#dependency-resolution) targeting `fabricloader`. 
+The mods declared `depends` and `breaks` clauses in the [fabric.mod.json](../loader/fabric-mod-json#dependency-resolution) targeting `fabricloader`.
 The minimum compatible Loader version will determine the compatibility level.
 
 If no Loader dependency is specified, the minimum compatibility level (equivalent to Fabric Loader 0.9.2) is used.
--- a/develop/loader/fabric-mod-json.md
+++ b/develop/loader/fabric-mod-json.md
@@ -231,7 +231,7 @@ The mods declared `depends` and `breaks` clauses in the [fabric.mod.json](../loa
 or `fabric-loader` (the former is preferred). The minimum compatible Loader version will determine
 the compatibility level.
 
-If no Loader dependency is specified, the minimum compatability level (equivalent to Fabric Loader 0.9.2) is used.
+If no Loader dependency is specified, the minimum compatibility level (equivalent to Fabric Loader 0.9.2) is used.
 
 **Note:** It is strongly encouraged to declare the minimum Fabric Loader dependency a mod is tested
 against!
@@ -244,7 +244,7 @@ For example, to enable enum extensions via mixin, at least this version of Loade
 }
 ```
 
-See the [compatability level documentation](../mixins/bytecode#compat-level) if any changes are applicable to your mod.
+See the [compatibility level documentation](../mixins/bytecode#compat-level) if any changes are applicable to your mod.
 
 :::
 
--- a/develop/loader/fabric-mod-json.md
+++ b/develop/loader/fabric-mod-json.md
@@ -222,6 +222,29 @@ The following keys will accept a dictionary of dependencies. For more details on
 - **`breaks`**: For mods whose together with yours might cause a game crash. **If any are present, Fabric Loader will trigger a crash**.
 - **`conflicts`**: For mods whose together with yours cause some kind of bugs, etc. For each conflicting mod present, Fabric Loader will log a warning.
 
+::: warning Dependencies Involving Fabric Loader
+
+Fabric Loader will examine a mod's dependency on Loader to control changes to mixin behavior that
+may be incompatible with earlier or newer versions of Loader.
+
+See the [compatability level documentation](../mixins/bytecode#compat-level) for more information
+on how Loader dependency is handled and if any changes are applicable to your mod.
+
+If no Loader dependency is declared, Loader will use the compatibility equivalent to version `0.9.2`.
+
+**Note:** It is strongly encouraged to declare the minimum Fabric Loader dependency a mod is tested
+against!
+
+For example, to enable enum extensions via mixin, at least this version of Loader is required:
+
+```json
+"depends": {
+   "fabricloader": ">=0.19.0"
+}
+```
+
+:::
+
 ### Semantic Versioning {#semantic-versioning}
 
 The key of each entry is the mod ID of the dependency.
--- a/develop/loader/fabric-mod-json.md
+++ b/develop/loader/fabric-mod-json.md
@@ -227,10 +227,11 @@ The following keys will accept a dictionary of dependencies. For more details on
 Fabric Loader will examine a mod's dependency on Loader to control changes to mixin behavior that
 may be incompatible with earlier or newer versions of Loader.
 
-See the [compatability level documentation](../mixins/bytecode#compat-level) for more information
-on how Loader dependency is handled and if any changes are applicable to your mod.
+The mods declared `depends` and `breaks` clauses in the [fabric.mod.json](../loader/fabric-mod-json#dependency-resolution) targeting `fabricloader`
+or `fabric-loader` (the former is preferred). The minimum compatible Loader version will determine
+the compatibility level.
 
-If no Loader dependency is declared, Loader will use the compatibility equivalent to version `0.9.2`.
+If no Loader dependency is specified, the minimum compatability level (equivalent to Fabric Loader 0.9.2) is used.
 
 **Note:** It is strongly encouraged to declare the minimum Fabric Loader dependency a mod is tested
 against!
@@ -242,6 +243,8 @@ For example, to enable enum extensions via mixin, at least this version of Loade
    "fabricloader": ">=0.19.0"
 }
 ```
+
+See the [compatability level documentation](../mixins/bytecode#compat-level) if any changes are applicable to your mod.
 
 :::
 
--- a/develop/mixins/bytecode.md
+++ b/develop/mixins/bytecode.md
@@ -604,27 +604,19 @@ Example:
 - Changes to local variable handling to better preserve local variables at the target location where previously they may have been incorrectly removed from Mixin's view
   - For more details, see [this mixin issue](https://github.com/SpongePowered/Mixin/issues/508).
 
-***
-
 #### Compatibility Level 0.14.0 (Fabric Loader 0.14.0, Mixin 0.8.6) {#compat-level-0140}
 
 - Gates a change to the filtering of `NEW` descriptors pursuant to a Mixin [bugfix](https://github.com/SpongePowered/Mixin/issues/515) allowing the constructor descriptor to be used to target an invocation.
 
-***
-
 #### Compatibility Level 0.16.5 (Fabric Loader 0.16.5, Mixin 0.8.7) {#compat-level-087}
 
 - `SHIFT` is now respected during injections.
-
-***
 
 #### Compatibility Level 0.17.0 (Fabric Loader 0.17.0, Mixin 0.8.7) {#compat-level-0170}
 
 - Local variable capture of method parameters
   - If possible, allow capturing by using the parameter's actual name
   - Otherwise, fallback to parameters named `arg<paramIndex>`, which previously was only the case when `argsOnly = false`. When `argsOnly` was `true`, they were named `arg<lvIndex>`
-
-***
 
 #### Compatibility Level 0.17.1 (Fabric Loader 0.17.1, Mixin 0.8.7) {#compat-level-0171}
 
--- a/develop/mixins/bytecode.md
+++ b/develop/mixins/bytecode.md
@@ -580,7 +580,7 @@ Here, the `name` parameter is being passed as a parameter into the lambda. Notic
 
 Mixin is a complicated and evolving tool that, for the sake of fixing bugs, sometimes changes behavior in such a way that working mods can break with a loader update.
 Starting with Fabric Loader 0.12.3 (a beta version, recommended to use 0.12.5), Loader is able to change mixin behavior on a per-mod basis, allowing old mods to continue working with
-the old behavior, and new mods to have access to any fixes they need. This is done by selecting the earliest Loader version compatible with a given mod, and then choosing the appropriate compat level.
+the old behavior, and new mods to have access to any fixes they need. New features may also be gated behind this block. This is done by selecting the earliest Loader version compatible with a given mod, and then choosing the appropriate compat level.
 
 Fabric Loader version compatibility is based on the mods declared `depends` and `breaks` clauses in the [fabric.mod.json](../loader/fabric-mod-json#dependency-resolution) targeting `fabricloader` or `fabric-loader` (the former is preferred). It is strongly encouraged to declare the
 minimum Fabric Loader dependency a mod is tested against, or if new functionality is desired, a newer loader version.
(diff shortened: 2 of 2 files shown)


develop/mixins/bytecode.md
```diff
@@ -574,3 +575,52 @@ static lambda$hello$1 (Ljava/lang/String;)V
...
+### Compatibility Level Determination {#compat-level-determination}
+
+### Changes {#compat-changes}
+
+#### Compatibility Level 0.10.0 (Fabric Loader 0.10.0, Mixin 0.8.4) {#compat-level-0100}
+
+- Changes to local variable handling to better preserve local variables at the target location where previously they may have been incorrectly removed from Mixin's view
+  - For more details, see [this mixin issue](https://github.com/SpongePowered/Mixin/issues/508).
+
+#### Compatibility Level 0.14.0 (Fabric Loader 0.14.0, Mixin 0.8.6) {#compat-level-0140}
```
MildestToucan: While not directly a compat level, perhaps it could also be noteworthy to note which versions start bundling important versions of MixinExtras, so here 0.15.0 being the minimum version to guarantee ME is present could be relevant.

develop/mixins/bytecode.md
```diff
@@ -574,3 +575,52 @@ static lambda$hello$1 (Ljava/lang/String;)V
...
+"depends": {
+   "fabricloader": ">=0.14.0"
+}
+```
+
+### Compatibility Level Determination {#compat-level-determination}
+
+### Changes {#compat-changes}
+
+#### Compatibility Level 0.10.0 (Fabric Loader 0.10.0, Mixin 0.8.4) {#compat-level-0100}
```
MildestToucan: A big issue here is that the compat level number actually corresponds to the **Mixin version**, not the Loader version.

As seen at:
https://github.com/FabricMC/fabric-loader/blob/b907c5b292fc062d75b6d8bf8255ac200109b992/src/main/java/net/fabricmc/loader/impl/launch/FabricMixinVersions.java#L33-L42

where the loader versions are associated with their Fabric Mixin compat levels.

the compatibility level `COMPATIBILITY_0_10_0` corresponds to Fabric Mixin version `0.10.0+mixin.0.8.4`, you can also see the corresponding Fabric Mixin versions in the [FabricUtil class](https://github.com/FabricMC/Mixin/blob/main/src/main/java/org/spongepowered/asm/mixin/FabricUtil.java)'s source code in comments next to the different constants.

So for compat level 0.10.0, it is more accurately Fabric Loader `0.12.0-` and Fabric Mixin 0.10.0+mixin.0.8.4

The same issue exists with the rest of the headers.

develop/mixins/bytecode.md
```diff
@@ -574,3 +575,52 @@ static lambda$hello$1 (Ljava/lang/String;)V
...
+
+- `SHIFT` is now respected during injections.
+
+#### Compatibility Level 0.17.0 (Fabric Loader 0.17.0, Mixin 0.8.7) {#compat-level-0170}
+
+- Local variable capture of method parameters
+  - If possible, allow capturing by using the parameter's actual name
+  - Otherwise, fallback to parameters named `arg<paramIndex>`, which previously was only the case when `argsOnly = false`. When `argsOnly` was `true`, they were named `arg<lvIndex>`
+
+#### Compatibility Level 0.17.1 (Fabric Loader 0.17.1, Mixin 0.8.7) {#compat-level-0171}
```
MildestToucan: As mentioned in my other comment, the header is mixing up loader and Mixin versions for the compat levels, but beyond that, this is critically lacking the Mixin enum extension feature, which is dependent on compat level 0.17.1, which on Fabric is Loader 0.19.0

develop/mixins/bytecode.md
```diff
@@ -574,3 +575,52 @@ static lambda$hello$1 (Ljava/lang/String;)V
 :::
 
 Here, the `name` parameter is being passed as a parameter into the lambda. Notice also how string concatenation is implemented with `invokedynamic`.
+
+## Differences in Mixin Bytecode Handling {#compat-level}
+
+Mixin is a complicated and evolving tool that, for the sake of fixing bugs, sometimes changes behavior in such a way that working mods can break with a loader update.
```
MildestToucan, line 581: This should also mention that sometimes *features* may be version gated, not just breaking changes in the form of bug fixing.

MildestToucan (changes requested):
Some mixups need to be resolved before this can be merged, and a couple things are missing.

develop/mixins/bytecode.md
dexman545: MixinExtras is bundled as a mod, so mods can explicitly depend on the version they need, or bundle a newer version than what floader ships. I don't believe ME gates anything behind a compat version asside from the already documented args handling.

Prior to floader 0.15, mods can (and frequently did) bundle ME. A mod requiring an earlier compat level does not change the version of ME used or its features.

I don't believe ME inclusion belongs in this section.

MildestToucan: Fair enough.

develop/mixins/bytecode.md
```diff
@@ -574,3 +575,53 @@ static lambda$hello$1 (Ljava/lang/String;)V
...
+
+```json
+"depends": {
+   "fabricloader": ">=0.14.0"
+}
+```
+
+### Compatibility Level Determination {#compat-level-determination}
+
+### Changes {#compat-changes}
```
its-miroma, line 587: I really don't want to come across as picky or difficult, but this kind of content is not in harmony with the rest of the docs.

Until now they haven't been documenting history, rather they showcased the latest version. Such a table is a better fit for a changelog, but ain't no nobody gonna be reading that lol.

However, I have to concede that if we ever want to complete the import of the wiki, that this rule must be relaxed. In fact, it's already been broken within the classtweaker pages.

We should maybe settle for this rule: Minecraft-related stuff is bound by the global version selector, but other stuff *must* specify a minimum version. The "other stuff" includes, but is not limited to:
- Fabric API? Maybe not?
  - What about backports of newer features?
- Mixin
- Loader
- CT
- ???

What do we think? Should we always specify version constraints in the docs?

develop/mixins/bytecode.md
```diff
@@ -574,3 +575,53 @@ static lambda$hello$1 (Ljava/lang/String;)V
...
+
+If no Loader dependency is specified, the minimum compatability level (equivalent to Fabric Loader 0.9.2) is used.
+
+Example:
+
+```json
+"depends": {
+   "fabricloader": ">=0.14.0"
+}
+```
```
its-miroma: This part should be in the `fabric.mod.json` page imo. Including the default value.

its-miroma (changes requested):
one comment is a review of the content, the other is a question that I'd like us to discuss.

develop/mixins/bytecode.md
dexman545: Unfortunately, these mixin changes do not have the best documentation. There's some minor mentions in the old Fabric blog posts, or the commit messages but that is really all I could find. A couple had a PR with very brief descriptions.

The compat levels do belong in the docs as there really isn't anywhere to find them and they function a bit like a magic number. Having them obviously leads to the question of "well, what changed to require a compat level?" for the reader, hence the list of changes. Most of these changes could be summarized as "lvt handling changed," which would probably be sufficient for most readers to at least know what to check in their mods when updating. 

An exhaustive description and comparison of the changes is likely unneeded here, but something for quick reference when updating a mod is probably beneficial.

I don't have any thoughts on version gating, but floader is meant to be MC version agnostic. If the docs say to use the latest floader (which I would agree with), well, these compat levels are part of latest floader and control it's behavior just like passing a system property or other API would so would still apply.

develop/mixins/bytecode.md
dexman545: A mention of the floader dep doing more than just a version check makes sense there, but the full description of how the level is chosen doesn't really fit to me.

develop/mixins/bytecode.md
its-miroma: yeah absolutely, I completely understand your perspective. I'm not saying this is not needed, I'm asking whether something needs to change in the docs paradigm.

recommending the latest version is a good instinct imo, but this doesn't excuse the fact that there is no changelog. this PR is trying to address that :+1:

develop/mixins/bytecode.md
its-miroma: > the full description of how the level is chosen doesn't really fit to me

why not? isn't it just: if specified, then that minimum version, otherwise default to X.

dexman545: It is not the specified version, it is the minimum compatible version specified (taking into account `breaks` and `depends`).

MildestToucan:
Something I wanted to mention: is the content really "too short" for its own page? It's a topic that will continuously keep expanding in content as Loader and Fabric Mixin get more updates.

Additionally, if it had to be somewhere other than its own page, is the bytecode page really the best place? It has nothing to do with the rest of the page, there's already Mixin subsections in the Loader introduction and fabric.mod.json articles.

If this ends up being merged before the Mixin intro docs I'll probably need to readjust some of the overlapping and possibly move the information to the added Mixin pages anyway. In the meantime there's no reason a reader would look for this information in the bytecode page, which I would argue is a good indicator that it doesn't belong there.

develop/mixins/bytecode.md
its-miroma: that is still a version resolution issue, not a mixin-related issue, so my opinion is that it should be in fabric-mod-json.

dexman545:
Most of the levels deal with how mixin handles the LVT, which seems more relevant to the bytecode page to me. If there was more general mixin documentation, I would agree it should go there.

develop/mixins/bytecode.md
dexman545: Moved

MildestToucan:
> Most of the levels deal with how mixin handles the LVT, which seems more relevant to the bytecode page to me. If there was more general mixin documentation, I would agree it should go there.

Fair enough. I'll make sure to look into possibly relocating it as part #680 or a follow-up PR when the time comes, making the reasonable assumption this gets merged first.

develop/loader/fabric-mod-json.md
```diff
@@ -222,6 +222,32 @@ The following keys will accept a dictionary of dependencies. For more details on
...
+::: warning Dependencies Involving Fabric Loader
+
+Fabric Loader will examine a mod's dependency on Loader to control changes to mixin behavior that
+may be incompatible with earlier or newer versions of Loader.
+
+The mods declared `depends` and `breaks` clauses in the [fabric.mod.json](../loader/fabric-mod-json#dependency-resolution) targeting `fabricloader`
+or `fabric-loader` (the former is preferred). The minimum compatible Loader version will determine
+the compatibility level.
+
+If no Loader dependency is specified, the minimum compatability level (equivalent to Fabric Loader 0.9.2) is used.
```
cassiancc:
```suggestion
If no Loader dependency is specified, the minimum compatibility level (equivalent to Fabric Loader 0.9.2) is used.
```

develop/loader/fabric-mod-json.md
```diff
@@ -222,6 +222,32 @@ The following keys will accept a dictionary of dependencies. For more details on
...
+
+For example, to enable enum extensions via mixin, at least this version of Loader is required:
+
+```json
+"depends": {
+   "fabricloader": ">=0.19.0"
+}
+```
+
+See the [compatability level documentation](../mixins/bytecode#compat-level) if any changes are applicable to your mod.
```
cassiancc:
```suggestion
See the [compatibility level documentation](../mixins/bytecode#compat-level) if any changes are applicable to your mod.
```

develop/loader/fabric-mod-json.md
```diff
@@ -222,6 +222,32 @@ The following keys will accept a dictionary of dependencies. For more details on
 - **`breaks`**: For mods whose together with yours might cause a game crash. **If any are present, Fabric Loader will trigger a crash**.
 - **`conflicts`**: For mods whose together with yours cause some kind of bugs, etc. For each conflicting mod present, Fabric Loader will log a warning.
 
+::: warning Dependencies Involving Fabric Loader
+
+Fabric Loader will examine a mod's dependency on Loader to control changes to mixin behavior that
+may be incompatible with earlier or newer versions of Loader.
+
+The mods declared `depends` and `breaks` clauses in the [fabric.mod.json](../loader/fabric-mod-json#dependency-resolution) targeting `fabricloader`
+or `fabric-loader` (the former is preferred). The minimum compatible Loader version will determine
```
cassiancc: Should we even mention `fabric-loader` existing at all? The example mod provides a dependency on `fabricloader`, and I don't believe I've seen that name in the wild at all.

cassiancc (changes requested):

develop/loader/fabric-mod-json.md
dexman545: `fabricloader` should be preferred since that is the [id of loader](https://github.com/FabricMC/fabric-loader/blob/master/src/main/resources/fabric.mod.json#L3), but for mixin compat floader [checks both ids](https://github.com/FabricMC/fabric-loader/blob/master/src/main/java/net/fabricmc/loader/impl/launch/FabricMixinBootstrap.java#L134).

Only [one mod on github seems to be using the dash](https://github.com/search?q=%5C%22fabric-loader%5C%22+language%3AJSON+path%3A**%2Ffabric.mod.json+NOT+is%3Afork&type=code) version. Can't easily check mods that are generating the json though

dexman545: The - version might be a holdover from some previous inconsistency. One error message uses it, as do some object shares.

<https://github.com/search?q=repo%3AFabricMC%2Ffabric-loader+fabric-loader+language%3AJava&type=code&l=Java>

its-miroma: Let's not mention it, I agree

develop/mixins/bytecode.md
```diff
@@ -574,3 +575,40 @@ static lambda$hello$1 (Ljava/lang/String;)V
...
+
+Mixin is a complicated and evolving tool that, for the sake of fixing bugs, sometimes changes behavior in such a way that working mods can break with a loader update.
+Starting with Fabric Loader 0.12.3 (a beta version, recommended to use 0.12.5), Loader is able to change mixin behavior on a per-mod basis, allowing old mods to continue working with
+the old behavior, and new mods to have access to any fixes they need. New features may also be gated behind this block. This is done by selecting the earliest Loader version compatible with a given mod, and then choosing the appropriate compat level.
+
+See [Compatibility Level Determination](../loader/fabric-mod-json#dependency-resolution) for more details on which compatibility level is selected.
+
+### Changes {#compat-changes}
+
+#### Compatibility Level 0.10.0 (Fabric Loader 0.12.0, Mixin 0.8.4) {#compat-level-0100}
```
MildestToucan: I feel like the header format is still a bit confusing currently, specifically regarding the "Mixin x.x.x" part;

Fabric Mixin does have an underlying upstream Mixin version as part of its versioning, e.g. `0.17.3+mixin.0.8.7`.

But with the header, while the compat level itself does match the first part of that versioning, I don't know if readers would be able to understand it properly.

For instance, with the header `Compatibility Level 0.10.0 (... Mixin 0.8.4)`, I don't know whether a reader would be able to properly parse it as "Compatibility level for Fabric Mixin `0.10.0+mixin.0.8.4`".

If someone isn't already familiar with Fabric Mixin versioning, and with the fact the compatibility level itself corresponds to the first half of a Fabric Mixin version, I fear it'd be more confusing than anything.

I'm unsure on how exactly the headers should be formatted instead. Maybe it'd be sufficient to just change `Mixin x.x.x` to the full Fabric Mixin version, even if it's a little redundant since the first half should match the compat level's number.

dexman545: It might make more sense to remove the mixin version, as it isn't all that relevant to the compat levels. Arguably, the compat level isn't strictly needed either since outside of looking it up from the mixin service it isn't directly applicable to users

So `Compatibility Level 0.10.0 (Fabric Loader 0.12.0)` or `Compatibility Level for Fabric Loader 0.12.0`

MildestToucan: Even if some more technical users wanted to look them up it could work to link to the Fabric Mixin class that lists the different compat levels somewhere so that they could just click that and look at the code there.

I think it might still be alright to keep the compatibility level version number itself, but the detailed Mixin version is probably unnecessary.

dexman545: Removed the mixin version. Not sure about linking to the code, do other docs do that? It's not terribly hard to find either.

MildestToucan: It wouldn't be too big of a deal to omit it, but it's not unprecedented to link back to relevant code for users that may be interested in some technical specs or details that aren't worth fully listing out in the general-user-oriented docs.

With the Mixin docs there's probably gonna be at least a couple of areas where I plan to delegate to source code, e.g. linking to the `MixinConfig` source code for more technical users, and only keeping the mixins.json page itself to the config options most relevant to mod development.

Earthcomputer (changes requested):
This does *not* belong in the bytecode page. It would be best as its own page. Moreover, this reads more like a changelog of implementation details than documentation. I would think about what a user actually needs to know to be able to use Mixin. They need to know that under specific circumstances they need to bump the loader dependency in their `fabric.mod.json`. I would document this in the relevant pages in warning blocks, rather than all in one place. A changelog belongs on Mixin's releases page.

cputnam-a11y:
> This does _not_ belong in the bytecode page. It would be best as its own page. Moreover, this reads more like a changelog of implementation details than documentation. I would think about what a user actually needs to know to be able to use Mixin. They need to know that under specific circumstances they need to bump the loader dependency in their `fabric.mod.json`. I would document this in the relevant pages in warning blocks, rather than all in one place. A changelog belongs on Mixin's releases page.

I will present that I have had several experienced developers in the support channels already express appreciation for having a quick cheat sheet all in one place. I would agree that each piece should be documented where it is used, but perhaps there is a benefit to having a quick reference in the mixin section somewhere on it's own page

Earthcomputer:
Is mixin compat versions a cheat sheet that is really needed? I'm not so sure. Even if it is, it needs to not be in this page, it needs to be in its own page.

Earthcomputer:
Would it not best be documented in the Mixin Javadocs themselves, in this class?
https://github.com/FabricMC/Mixin/blob/main/src/main/java/org/spongepowered/asm/mixin/FabricUtil.java

Someone should go and add to the Javadocs of each of those fields explaining what the compat level changes.

its-miroma:
I agree with Earth, the Mixin repo needs to have a changelog. Documenting changes in the compat level class is also useful, but less necessary if there's a changelog. Also, the table in that README is out of date.

dexman545:
Most of the compat levels seem to deal with how the LVT is handled, so to me is most relevant to the bytecode page. There isn't enough information to receive it's own dedicated page. I would be fine with this being closed in favor of the mixin docs PR that includes this information.

Earthcomputer:
> Most of the compat levels seem to deal with how the LVT is handled

This is only half-true even right now, and will get less true in the future.

PR #693, opened 2026-08-03
Document mod discovery and resolution

dexman545:


develop/loader/mod-resolution.md
```diff
@@ -0,0 +1,161 @@
...
+### Built-in Mods {#built-in-sources}
+
+For minecraft, Loader provides several builtin mods:
+
+- `minecraft`
+- `java`
+  - The java version being used to run the game, such as `8` or `25`.
+
+Fabric Loader will also list itself in the mod list as `fabricloader`, but it is technically not a
+builtin mod.
```
its-miroma, line 172: sure... is `minecraft` or `java` a built-in mod though? I don't think this line is needed.

Also, again, are these names official or did you choose them for this page?

develop/loader/mod-resolution.md
```diff
@@ -0,0 +1,161 @@
...
+choosing the most compatible version of each mod. Only one version of a mod can be loaded at a time.
+
+## Mod Discovery {#mod-discovery}
+
+To choose which mods to load, it must first find all the mods, such as all the mods in the `mods`
+folder. See [Mod Sources](#mod-sources) for other sources. As mods can contain other mods
+(see [Loom's `include` configuration](../loom/options)), Loader will search within each mod for more
+mods. Note that bundled mods can also bundle mods of their own. There are three types of mods:
+
+### Plain Mods {#plain-mods-disc}
```
its-miroma, line 18: Is "Plain Mod" actually the term used usually? I'm curious

develop/loader/mod-resolution.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Plain Mods {#plain-mods-res}
+
+Plain mods are generally preferred over nested mods, and some version of this mod shall be loaded.
+
+### Nested Mods {#nested-mods-res}
+
+Nested mods are not required to be loaded, unless they are required by another mod. This is useful
+for mods that provide additional features to other mods that may not be present, or allowing the
+parent mod to provide different implementations depending on the environment or other loaded mods.
```
its-miroma: Let me see if I understand this correctly: If example-mod includes a nested compat-mod for compatibility with another-mod, it can be configured to only load if another-mod is also available? I'm guessing so, by adding another-mod to compat-mod's depends.

What are the advantages of using a nested mod, instead of having all the logic in the main mod? Is it just to minimize the amount of code loaded unnecessarily, or just a refactoring preference?

develop/loader/mod-resolution.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Builtin Mods {#builtin-mods-res}
+
+These mods are always loaded.
+
+### Examples {#examples}
```
its-miroma, line 68: Do you think flowchart support would be useful for these examples, to show the logic?

develop/loader/mod-resolution.md
```diff
@@ -0,0 +1,161 @@
...
+- `minecraft`
+- `java`
+  - The java version being used to run the game, such as `8` or `25`.
+
+Fabric Loader will also list itself in the mod list as `fabricloader`, but it is technically not a
+builtin mod.
+
+### Mods Folder {#mods-folder}
+
+By default, Loader will look for mods in the `mods` folder located in the `minecraft` folder.
```
its-miroma:
```suggestion
By default, Loader will look for mods in the `mods` folder located in the `.minecraft` folder.
```

develop/loader/mod-resolution.md
```diff
@@ -0,0 +1,161 @@
...
+This search is not recursive, so it will not look in subfolders.
+
+This folder can be changed by setting the `fabric.modFolder` system property.
```
its-miroma: Where?

its-miroma (changes requested):
superficial review

develop/loader/mod-resolution.md
CelDaemon: This is really highly launcher dependent, can we just refer to it as the instance folder or something?

develop/loader/mod-resolution.md
CelDaemon: It's a system property, something you can set with jvm flags `-Dfabric.modFolder=somedir`. Where to set it depends on how you're running the game.

develop/loader/mod-resolution.md
its-miroma: My understanding is that we assume the vanilla launcher in the Docs. I think "`.minecraft` installation folder" is fine, or even "`.minecraft` instance folder"

CelDaemon: Yeah sure, seems fine

develop/loader/mod-resolution.md
dexman545: These terms come from floader, so I felt it best to stick with them. Personally, I have only used the terms builtin and JiJed (for nested), with plain mods just being "mods" with no differentiator.

https://github.com/FabricMC/fabric-loader/blob/master/src/main/java/net/fabricmc/loader/impl/discovery/ModCandidateImpl.java#L65-L79

develop/loader/mod-resolution.md
dexman545: Yes, Minecraft and java are both builtin mods provided by the game provider. Floader is always loaded and present though, but is not marked as a built in mod. 

See other comment for where the terms come from.

https://github.com/FabricMC/fabric-loader/blob/master/minecraft/src/main/java/net/fabricmc/loader/impl/game/minecraft/MinecraftGameProvider.java#L118-L133

https://github.com/FabricMC/fabric-loader/blob/master/src/main/java/net/fabricmc/loader/impl/discovery/ModCandidateImpl.java#L152-L154

develop/loader/mod-resolution.md
dexman545: Yes, these can usually be set by the launcher. Perhaps at a later date the installing mods player guide can be updated with instructions on how to set these.

develop/loader/mod-resolution.md
dexman545: Yes, nested mods only load if they can (dependencies are ok). It is only an error to not load a nested mod when it is required by another mod that depends on that nested mod and that mod has to load (such as a plain mod).

The same can be done to disable a compat mod when another mod conflicts with it.

It is possible to gate compatibility behind manual checks in a mod, this optional loading of dependencies can have an easier time of setting that up (no fear of accidentally referencing the wrong class), and floader is better able to handle more complex version requirements than "is mod b loaded." 

The other advantage is with classtweakers and mixins. It is possible to setup mixins to only apply based on loaded mods with a mixinconfigplugin, but I don't believe classtweakers have a way to optionally load (though they do ignore missing entries).

Personally, I see their greatest use with providing different implementations for different MC versions, as it is much easier to setup and use than something like stonecutter or using reflection.

Player would probably be able to list other advantages/differences - this isn't a system I use frequently, and finding examples of this in the wild is hard as it hasn't really been documented before.

develop/loader/mod-resolution.md
dexman545: Maybe? Not sure on walking through the logic step by step as it is a pretty complicated SAT solver handling all the constraints at once, unless you mean something simplified like "aa breaks on b, so it is disabled"

Something to more easily show the  nested mods would be nice.

develop/loader/mod-resolution.md
dexman545: Expanded this section on the differences and added an example of it affecting mod loading.

develop/loader/mod-resolution.md
```diff
@@ -0,0 +1,196 @@
+---
+title: Mod Resolution
+description: An explanation of how Fabric Loader chooses which mods to load.
+authors:
+  - Deximus-Maximus
```
its-miroma: This needs to be your GH username.

```suggestion
  - dexman545
```

develop/loader/mod-resolution.md
```diff
@@ -0,0 +1,196 @@
...
+resources:
```
its-miroma: Did you want to add any?

its-miroma (changes requested):

Issue #695, opened 2026-08-05
Workstations page: incorrect slot index in slotsChanged sync packet, and inconsistent slot ordering with quickMoveStack section


Syhmac:
### Page

https://docs.fabricmc.net/develop/blocks/workstations

### Issue

In the "Creating a Menu" section, `UpgradingMenu`'s constructor adds slots in this order:

1. `Slot(this.input, 0, 27, 47)` — global slot index 0
2. `Slot(this.input, 1, 76, 47)` — global slot index 1
3. `UpgradingResultSlot(this, this.output, 0, 134, 47)` — global slot index **2**

However, in `slotsChanged()`, the manual sync packet is sent for slot index **0**:

```java
this.setRemoteSlot(0, result);
((ServerPlayer) this.player).connection.send(new ClientboundContainerSetSlotPacket(this.containerId, this.incrementStateId(), 0, result));
```

Since the result slot is actually at global index 2 (not 0), this sends a `ClientboundContainerSetSlotPacket` for the **first input slot** instead of the result slot. In practice this causes the first input slot's client-side item to flicker/disappear whenever `slotsChanged` fires (e.g. on any input change, or after `onTake` modifies the input), because the client gets told "slot 0 = result" (often `ItemStack.EMPTY`), overwriting the correct rendering of the first input slot until the next full `broadcastChanges()` sync happens (e.g. from an unrelated slot click).

### Additional inconsistency

Further down, in "Implementing `quickMoveStack`", the constants assume the opposite slot order:

```java
private static final int RESULT_SLOT = 0;
private static final int INPUT_SLOTS_START = RESULT_SLOT + 1; // 1
```

This treats the result slot as index 0 and inputs as starting at index 1 — contradicting the slot order established in the constructor earlier on the same page.

### Suggested fix

Either:
- Reorder the `addSlot` calls in the constructor to add the result slot first (matching the `quickMoveStack` section), and fix the packet's hardcoded index accordingly, **or**
- Remove the manual `setRemoteSlot`/`ClientboundContainerSetSlotPacket` call entirely and rely on `AbstractContainerMenu`'s automatic `broadcastChanges()`, which handles slot indices correctly without hardcoding — this seems like the safer recommendation for a tutorial, since it avoids the whole class of bugs where copying this pattern into a different slot layout silently breaks.

Either way, the constructor's slot order and the `quickMoveStack` constants should be made consistent with each other.

### Context

[...]

PR #702, opened 2026-08-13
Cleanup upgrading menu

CelDaemon:
Fixes half of #695

The quick move code from SuperiorUpgradingMenu is simply incorrect, and will need to be rewritten + moved into the normal menu. It may also be worth looking into writing tests for these things, with how brittle they are.

### DIFF
--- a/reference/latest/src/main/java/com/example/docs/menu/custom/UpgradingMenu.java
+++ b/reference/latest/src/main/java/com/example/docs/menu/custom/UpgradingMenu.java
@@ -3,11 +3,7 @@ package com.example.docs.menu.custom;
 import java.util.List;
 import java.util.Optional;
 
-import org.jspecify.annotations.Nullable;
-
-import net.minecraft.network.protocol.game.ClientboundContainerSetSlotPacket;
 import net.minecraft.server.level.ServerLevel;
-import net.minecraft.server.level.ServerPlayer;
 import net.minecraft.world.Container;
 import net.minecraft.world.SimpleContainer;
 import net.minecraft.world.entity.player.Inventory;
@@ -18,6 +14,7 @@ import net.minecraft.world.inventory.ResultContainer;
 import net.minecraft.world.inventory.Slot;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.item.crafting.RecipeHolder;
+import net.minecraft.world.level.Level;
 
 import com.example.docs.block.ModBlocks;
 import com.example.docs.menu.ModMenuTypes;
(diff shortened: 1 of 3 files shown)


reference/latest/src/main/java/com/example/docs/menu/custom/UpgradingMenu.java
```diff
@@ -74,41 +66,34 @@ protected void onTake(final Player player, final ItemStack stack) {
...
-					RecipeHolder<UpgradingRecipe> recipeHolder = maybeRecipe.get();
-					UpgradingRecipe recipe = recipeHolder.value();
-
-					if (this.output.setRecipeUsed((ServerPlayer) this.player, recipeHolder)) {
-						ItemStack recipeResult = recipe.assemble(recipeInput);
-
-						if (recipeResult.isItemEnabled(level.enabledFeatures())) {
-							result = recipeResult;
-						}
-					}
```
its-miroma, line 93: You dropped these checks, are they redundant or unnecessary?

reference/latest/src/main/java/com/example/docs/menu/custom/UpgradingMenu.java
```diff
@@ -74,41 +66,34 @@ protected void onTake(final Player player, final ItemStack stack) {
...
-				}
-
-				this.output.setItem(0, result);
-
-				/*
-				Alternatively, call broadcastChanges instead of setting the remote slot and sending a packet.
-				Based on how your Menu is structured, you may not need to manually call any syncing method, but it is recommended that you are very sure of yourself before you remove these calls to avoid server-client desyncs.
-				 */
-				this.setRemoteSlot(0, result);
-				((ServerPlayer) this.player).connection.send(new ClientboundContainerSetSlotPacket(this.containerId, this.incrementStateId(), 0, result));
```
its-miroma, line 107: Is this unnecessary?

its-miroma (commented):
Just a few questions.

Also, does the page need any changes?

reference/latest/src/main/java/com/example/docs/menu/custom/UpgradingMenu.java
CelDaemon: Honestly I'm not sure. This code is based on `SmithingMenu`, which doesn't seem to bother with checking for this at all.

reference/latest/src/main/java/com/example/docs/menu/custom/UpgradingMenu.java
CelDaemon: Seemingly, the menu works correctly without them. And `SmithingMenu` also does not do anything like this.

cassiancc (approved):

reference/latest/src/main/java/com/example/docs/menu/custom/UpgradingMenu.java
CelDaemon: I think it's in our best interest to drop these checks for simplicity.

reference/latest/src/main/java/com/example/docs/menu/custom/UpgradingMenu.java
CelDaemon: I have thoroughly tested in both singleplayer and dedicated server. All interactions with the slot (except for quick moves, which were already broken), seem to work correctly.

cassiancc:
Page doesn't seem to need any changes - going to go ahead and merge this.

Issue #710, opened 2026-08-20
Lessons from HayesGaze's video


its-miroma:
Just a bunch of ideas I jotted down after watching Hayes' video:
<https://www.youtube.com/watch?v=RZb-3PIhz74>

- Should we standardise the Prerequisites sections? Maybe as checklists?
- Sections of the Docs, especially the Creating Your Firsts pages, should dedicate more space to explaining *why* we're doing things. Create this, add this, use this helper, does not allow the user to understand the reason for these classes.
- We still don't have a page/definition of registries (:
- We should mention that we do not include imports in the code blocks - both in cgl and in the introduction - because they are detected by the IDE or they can be found in the reference mod files.
- We should advise modders to be careful with old posts / code / tutorials, because of the many changes to Minecraft over the last few years (deobfuscation, rendering, data-driving, etc)
- Maybe we could recommend some trusted video tutorials? Obv we would label them as third-party, and we would ask for permission, but for example Kaupenjoe's playlist comes to mind.
- "WHY ARE THEY CHANGING IT AGAIN" :sob::v:
- I was thinking of repurposing https://fabric-docs.netlify.app/ (currently fully broken lol) as a stable URL for the previews of upcoming versions. In other words, when we run the bump script to bump to 26.3, that URL should link to the 26.3 WIP pages (with a clear banner). That way, we can have the WIP versions linked to in prod. Hayes recorded their video on the 21st of June, when Cassian's port/26.2 already existed, but they never found out. And they wouldn't have been able to until a month later.
- Maybe, as a suggestion for new modders, we could ask them to avoid the latest version if we haven't even merged it yet to the docs. Especially with how quickly Mojang is changing things, we could add a disclaimer to "use a more established version" if you are new to modding. I'm not saying 1.21.11, I'm just saying: don't use 26.2 5 days after release.
- We need a better way of explaining the folder structure. I need to study whether there exists already some kind of md-it plugin, or vitepress plugin, for that, or if I should code (or vibecode) something for that. "Starlight" apparently offers something similar that I like: https://starlight.astro.build/components/file-tree/
- Guidite is a pun on a material for the guide, but maybe we can change that for clarity (related: #562)
- We need to find a way to let modders understand that they can change the IDs of things when they add them. Just look at this: https://github.com/search?q=guidite_&type=code
- Seems like the page on Events does not explain their sidedness (related: #230)
[...]

Issue #707, opened 2026-08-19
Creating your first entity tutorial is missing some information


markaull:
I just completed the "Creating Your First Entity" part of the Fabric documentation. The following things were unclear or missing from the documentation:
1. After using the Template mod generator, files must be in the correct directories:
\src\client\java\com\example\docs\client\mixin\ExampleClientMixin.java
\src\client\java\com\example\docs\entity\ExampleModCustomEntityClient.java
\src\client\java\com\example\docs\entity\animation\MiniGolemAnimations.java
\src\client\java\com\example\docs\entity\client\EntityClient.java
\src\client\java\com\example\docs\entity\client\mixin\ExampleClientMixin.java
\src\client\java\com\example\docs\entity\model\MiniGolemEntityModel.java
\src\client\java\com\example\docs\entity\model\ModEntityModelLayers.java
\src\client\java\com\example\docs\entity\renderer\MiniGolemEntityRenderer.java
\src\client\java\com\example\docs\entity\state\MiniGolemEntityRenderState.java
\src\client\resources\entity.client.mixins.json
\src\client\resources\assets\example-mod\textures\entity\mini_golem_small.png
\src\main\java\com\example\docs\ExampleMod.java
\src\main\java\com\example\docs\entity\MiniGolemEntity.java
\src\main\java\com\example\docs\entity\ModEntityTypes.java
\src\main\java\com\example\docs\entity\attribute\EntityAttributesGameTest.java
\src\main\java\com\example\docs\entity\attribute\ExampleModAttributes.java
\src\main\java\com\example\docs\entity\attribute\ModAttributes.java
\src\main\java\com\example\docs\entity\mixin\ExampleMixin.java
\src\main\java\com\example\docs\mixin\ExampleMixin.java
\src\main\resources\entity.mixins.json
\src\main\resources\fabric.mod.json
\src\main\resources\assets\example-mod\icon.png

2. Mod ID must match between ExampleMod.java
public static final String MOD_ID = "example-mod";
and fabric.mod.json
"name": "example-mod",

3. Entry points must be set correctly in fabric.mod.json, template client file should be removed
	"entrypoints": {
		"main": [
			"com.example.docs.ExampleMod"
		],
		"client": [
			"com.example.docs.entity.ExampleModCustomEntityClient"
		]
	},

4. Code to register the entity must be added to the main entry point function (or a file for the main entry point should be provided):
ModEntityTypes.registerModEntityTypes();
[...]

Issue #608, opened 2026-05-14
Enable Null Checks


its-miroma:
Suggested-by: @CelDaemon

PR #714, opened 2026-08-26
Projectiles

dicedpixels:
More or less a port of [Creating a Custom Projectile](https://wiki.fabricmc.net/tutorial:projectiles).
https://deploy-preview-714--fabric-docs.netlify.app/develop/entities/projectiles


.vitepress/sidebars/develop.ts
```diff
@@ -170,6 +170,10 @@ export default [
         text: "develop.entities.damage_types",
         link: "/develop/entities/damage-types",
       },
+      {
```
cassiancc: Nit: I'd move this below "Creating Your First Entity", like how the Item docs are formatted.

develop/entities/projectiles.md
```diff
@@ -0,0 +1,147 @@
...
+  - onlyspxctre
+  - patrickmsm
+  - tianjun
+  - upcraftlp
+resources:
+  https://minecraft.wiki/w/Projectile: Projectiles - Minecraft Wiki
+  https://docs.neoforged.net/docs/entities/#projectiles: Projectiles - NeoForge Docs (except Neo exclusives)
+---
+
+Projectiles are entities that can be thrown or fired by players, other entities and dispensers. In this guide, we'll look into implementing a simple projectile like a snowball.
```
cassiancc: You call out dispensers here, but this doc doesn't add support for dispensing the projectile.

reference/latest/src/main/java/com/example/docs/projectile/ExampleModProjectile.java
```diff
@@ -0,0 +1,53 @@
...
+					HOT_TATER_ID,
+					EntityType.Builder.<HotTaterEntity>of(HotTaterEntity::new, MobCategory.MISC)
+									.sized(0.25F, 0.25F) // Hitbox width and height.
+									.clientTrackingRange(4) // How far (in chunks) clients see the entity.
+									.updateInterval(10) // Ticks between position updates sent to clients.
+									.build(ResourceKey.create(Registries.ENTITY_TYPE, HOT_TATER_ID)));
+	// #endregion register_entity
+
+	// #region register_item
+	public static final Item HOT_TATER_ITEM = Registry.register(
```
cassiancc: I believe we usually use `ModItems` register methods rather than manually calling `setId`.

develop/entities/projectiles.md
```diff
@@ -0,0 +1,147 @@
...
+
+We use the Minecraft-provided `ThrownItemRenderer` for our entity. This is the same renderer used by throwable items like the snowball.
+
+<<< @/reference/latest/src/client/java/com/example/docs/projectile/ExampleModProjectileClient.java#renderer
+
+At this point, you can obtain the projectile item in game and test out the functionality.
+
+## Finalizing the Item {#finalizing-the-item}
+
+Even though our projectile works, it still doesn't have a model, texture or a name. Follow the [Creating Your First Item](../items/first-item) guide to add these and finalize the item.
```
cassiancc: We usually provide the textures for custom items, see [Fluid Buckets](https://deploy-preview-714--fabric-docs.netlify.app/develop/fluids/first-fluid#buckets) for a good reference on how to format this when the item isn't the main focus of the article.

reference/latest/src/main/java/com/example/docs/projectile/HotTaterItem.java
```diff
@@ -0,0 +1,56 @@
...
+											@NonNull ItemStack itemStack,
+											@NonNull Direction direction) {
+		// Used when the item is shot from a dispenser instead of thrown by a player.
+		return new HotTaterEntity(level, position.x(), position.y(), position.z(), itemStack);
+	}
+	// #endregion as_projectile
+
+	// #region use
+	@Override
+	public @NonNull InteractionResult use(@NonNull Level level, @NonNull Player player, @NonNull InteractionHand hand) {
```
cassiancc: We ought to decide what to about these annotations. Minecraft doesn't use them, it uses `@NullMarked` on the entire class.

cassiancc (changes requested):

reference/latest/src/main/java/com/example/docs/projectile/HotTaterItem.java
cputnam-a11y: Does it make sense to move the package info generation into loom as a utility task and use that? Idea has support now for picking them up in generated source dirs

develop/entities/projectiles.md
dicedpixels: I removed mentions of dispensers. Perhaps we can think about expanding the guide later.

develop/entities/projectiles.md
dicedpixels: Following the same pattern now.

reference/latest/src/main/java/com/example/docs/projectile/ExampleModProjectile.java
dicedpixels: Completely forgot that. Now using `ModItems` utility method.

reference/latest/src/main/java/com/example/docs/projectile/HotTaterItem.java
dicedpixels: I've added `@NullMarked` and I do think it's better since it reduces a bit of code clutter.

cassiancc (dismissed):

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+  - haykam
+  - kanpov
+  - NetUserGet
+  - onlyspxctre
+  - patrickmsm
+  - tianjun
+  - upcraftlp
+resources:
+  https://minecraft.wiki/w/Projectile: Projectiles - Minecraft Wiki
+  https://docs.neoforged.net/docs/entities/#projectiles: Projectiles - NeoForge Docs (except Neo exclusives)
```
its-miroma: Which Neo exclusives are there even in that section?

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+---
+
+Projectiles are entities that can be thrown or fired by players or other entities. In this guide, we'll look into implementing a simple projectile like a snowball.
+
+We'll call our projectile a Hot Tater. It will be a potato that sets the block or entity it hits on fire.
```
its-miroma, line 23:
```suggestion
Projectiles are entities that can be thrown or fired by players or other entities.

In this guide, we'll look into implementing a simple projectile like a snowball. We'll call our projectile a Hot Tater - it will be a potato that sets the block or entity it hits on fire.
```

Can we make it give the entity it hits saturation as well? I just feel like this is too much of a clone of the fireball...

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Prerequisites {#prerequisites}
+
+Creating a projectile requires you to register an item as well as an entity. Therefore we suggest going through the [Creating Your First Item](../items/first-item) and [Creating Your First Entity](./first-entity) guides.
```
its-miroma:
```suggestion
Creating a projectile requires you to register an item as well as an entity, therefore we suggest going through the [Creating Your First Item](../items/first-item) and [Creating Your First Entity](./first-entity) guides.
```

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+resources:
+  https://minecraft.wiki/w/Projectile: Projectiles - Minecraft Wiki
+  https://docs.neoforged.net/docs/entities/#projectiles: Projectiles - NeoForge Docs (except Neo exclusives)
+---
+
+Projectiles are entities that can be thrown or fired by players or other entities. In this guide, we'll look into implementing a simple projectile like a snowball.
+
+We'll call our projectile a Hot Tater. It will be a potato that sets the block or entity it hits on fire.
+
+## Prerequisites {#prerequisites}
```
its-miroma: Food for thought: some pages use `::: info PREREQUISITES`, some use a heading like this, maybe we should standardize. Related: #710

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+
+Creating a projectile requires you to register an item as well as an entity. Therefore we suggest going through the [Creating Your First Item](../items/first-item) and [Creating Your First Entity](./first-entity) guides.
+
+## Creating the Projectile Entity {#creating-the-projectile-entity}
+
+Let's create `HotTaterEntity` by extending `ThrowableItemProjectile`. This class should be in your `main` source set.
```
its-miroma:
```suggestion
Let's create a `HotTaterEntity` by extending `ThrowableItemProjectile`. This class should be in your `main` source set.
```

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+
+The `ThrowableItemProjectile` class handles the physics logic and knows how to store the item form of the projectile.
```
its-miroma:
```suggestion
The `ThrowableItemProjectile` class handles the physics and the item form of the projectile.
```

reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java
```diff
@@ -0,0 +1,83 @@
...
+		super(type, level);
+	}
+
+	public HotTaterEntity(Level level, LivingEntity owner, ItemStack itemStack) {
+		super(ExampleModProjectile.HOT_TATER_ENTITY_TYPE, owner, level, itemStack);
+	}
+
+	public HotTaterEntity(Level level, double x, double y, double z, ItemStack itemStack) {
+		super(ExampleModProjectile.HOT_TATER_ENTITY_TYPE, x, y, z, level, itemStack);
+	}
```
its-miroma, line 31: I am confused by these constructors:

- Why do the second and the third use `HOT_TATER_ENTITY_TYPE`, while the first passes `type` as an argument?
- Why flip the argument order relative to vanilla (putting Level as the first argument)?
- I guess we could have used Position instead of x+y+z, were it not for vanilla's ctor?

reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java
```diff
@@ -0,0 +1,83 @@
...
+	// #endregion on_hit_entity
+
+	// #region on_hit
+	@Override
+	protected void onHit(HitResult hitResult) {
+		super.onHit(hitResult);
+
+		// Discard the projectile on any hit, or it sinks into the ground forever.
+		if (!level().isClientSide()) {
+			discard();
```
its-miroma: `this.`?

reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java
```diff
@@ -0,0 +1,83 @@
...
+import net.minecraft.world.item.Item;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.level.Level;
+import net.minecraft.world.level.block.BaseFireBlock;
+import net.minecraft.world.phys.BlockHitResult;
+import net.minecraft.world.phys.EntityHitResult;
+import net.minecraft.world.phys.HitResult;
+
+// #region entity
+@NullMarked
```
its-miroma, line 18: Do we include `@NullMarked` in the snippet? Also, we should totally use Null annotations everywhere (related: #608)

reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java
```diff
@@ -0,0 +1,83 @@
...
+	// #endregion default_item
+
+	// #region on_hit_block
+	@Override
+	protected void onHitBlock(BlockHitResult hitResult) {
+		super.onHitBlock(hitResult);
+		Level level = level();
+
+		// Only modify the world on the server.
+		if (!level.isClientSide()) {
```
its-miroma, line 48: Should we use early returns for code like this?

reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java
```diff
@@ -0,0 +1,83 @@
...
+			// If the projectile hits a block, place fire on the face it hit.
+			BlockPos pos = hitResult.getBlockPos().relative(hitResult.getDirection());
+
+			if (level.isEmptyBlock(pos)) {
```
its-miroma, line 52: When realistically is this check going to be hit? When the player is waterlogged? When the player that throws the Hot Tater is suffocating in a block?

sidebar_translations.json
```diff
@@ -39,6 +39,7 @@
   "develop.entities.effects": "Mob Effects",
   "develop.entities.damage_types": "Damage Types",
   "develop.entities.attributes": "Attributes",
+  "develop.entities.projectiles": "Projectiles",
```
its-miroma: Should it be "Custom Projectiles"? "Creating Your First Projectile"?

We need to uniformize the titles...

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+### Constructors {#constructors}
+
+We define 3 constructors. They're used by entity registration, projectile spawning and projectile conversion respectively.
+
+<<< @/reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java#constructors
+
+### Overrides {#projectile-entity-overrides}
+
+We will be overriding `getDefaultItem()`, `onHitBlock()`, `onHitEntity()` and `onHit()`.
+
```
its-miroma:
```suggestion
```

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+**`getDefaultItem()`**
```
its-miroma:
```suggestion
### Overriding `getDefaultItem()` {#override-get-default-item}
```

ditto for the other three. oh wait, maybe you didn't like that they are in the Table of Contents?

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+
+Defines the item form of this projectile.
+
+<<< @/reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java#default_item
+
+::: info
+
+We will register our item in a bit. You may also register the item first. See the [Registering the Item](#registering-the-item) section.
```
its-miroma:
```suggestion
Your IDE might tell you that it cannot resolve the item: we will create it soon, in the [Registering the Item](#registering-the-item) section.
```

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+
+:::
+
+**`onHitBlock()`**
+
+Defines the behavior when this projectile hits a block. We check where the projectile has hit and then set the top face of that block on fire. This logic is handled on the server side.
```
its-miroma: > and then set the top face of that block on fire.

If I understand the code correctly it sets the *hit* face on fire, not the top face.

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+
+<<< @/reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java#on_hit_block
+
+**`onHitEntity()`**
+
+Defines the behavior when this projectile hits an entity. We set the entity that was hit on fire for 5 seconds.
```
its-miroma, line 67: Again, as I said before: this is so similar to the Ghast Fireball and like the fireball from Bedwars. Maybe that's intentional, but maybe we can create something closer to a throwable potato (gives entities Saturation, and plants potato on farmland? dirt? idrk)

This doesn't matter much tho

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+
+<<< @/reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java#on_hit_entity
+
+**`onHit()`**
+
+Defines the behavior when this projectile hits anything, whether a block or an entity. Vanilla projectiles like the snowball discard themselves here so that the projectile is always removed on hit. Without this, a projectile that hits an entity would just keep going.
```
its-miroma:
```suggestion
Defines the behavior when this projectile hits anything, whether a block or an entity. We will use this to discard the projectile, so that it is removed on hit; without this, a projectile that hits an entity would just keep going.
```

I'm guessing that it would keep going if it hit a block too, right?

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+
+<<< @/reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java#on_hit
+
+## Creating the Item {#creating-the-item}
+
+We register a simple item. Since we need to implement the throwing logic, our class `HotTaterItem` will extend `Item` and implement `ProjectileItem`. This class should be in your `main` source set.
+
+<<< @/reference/latest/src/main/java/com/example/docs/projectile/HotTaterItem.java#item
+
+It's a simple item implementation with a standard constructor. But let's analyze the rest of the implementation.
```
its-miroma:
```suggestion
It's a standard item implementation, with some special methods from `ProjectileItem`. Let's look at them:
```

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Overrides {#item-overrides}
+
+We override `asProjectile()` and `use()`.
+
```
its-miroma:
```suggestion
```

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+**`asProjectile()`**
```
its-miroma:
```suggestion
### Overriding `asProjectile()` {#override-as-projectile}
```

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+
+This method converts the item into its entity form.
+
+<<< @/reference/latest/src/main/java/com/example/docs/projectile/HotTaterItem.java#as_projectile
+
+**`use()`**
```
its-miroma:
```suggestion
### Overriding `use()` {#pverride-use}
```

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+
+Defines the action that happens when the item is used. In our case, we use the `Projectile.spawnProjectileFromRotation()` utility method to spawn the projectile. In addition to the obvious parameters of the level, item stack, and player, this utility method takes three floats dictating the y-offset, power, and uncertainty.
```
its-miroma:
```suggestion
Defines the action that happens when the item is used. In our case, we call the `Projectile.spawnProjectileFromRotation()` utility method to spawn the projectile.

In addition to the standard parameters (level, item stack, and player), this utility method takes three additional floats:
```

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+
+- **Y-Offset**: Rotation in the y direction (looking up or down).
```
its-miroma, line 97: Should we use "pitch" from aerodynamics?

Pitch diagram
https://en.wikipedia.org/wiki/Aircraft_principal_axes

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+- **Power**: Scales how fast the projectile moves in blocks.
```
its-miroma, line 98: What does "in blocks" mean? "through blocks" maybe?

It's not a unit of measurement I would guess...


```suggestion
- **Power**: Speed multiplier for the projectile movement.
```

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+- **Uncertainty**: Scales the randomness in what direction the projectile goes when shot.
```
its-miroma:
```suggestion
- **Uncertainty**: Range of directions that the projectile can randomly go in when shot.
```

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+
+Finally, we consume one item from the stack and mark the interaction as successful.
```
its-miroma: Let's also mention the stat we award.

reference/latest/src/main/java/com/example/docs/projectile/HotTaterItem.java
```diff
@@ -0,0 +1,58 @@
...
+		// Spawn the projectile on the server only.
+		if (level instanceof ServerLevel serverLevel) {
+			// Y-offset = 0, power = 1.5, uncertainty = 1.0.
+			Projectile.spawnProjectileFromRotation(HotTaterEntity::new,
+					serverLevel,
+					itemStack,
+					player,
+					0.0F,
+					1.5F,
+					1.0F);
```
its-miroma: Consider adding the names of the parameters as comments (`/* uncertainty: */ 1.0F`)

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+
+<<< @/reference/latest/src/main/java/com/example/docs/projectile/HotTaterItem.java#use
+
+## Entity Renderer {#entity-renderer}
+
+Like most entities, your projectile will also need a renderer to render its visuals while it's on the move. For this guide, we'll be using Minecraft's built-in `ThrownItemRenderer`. If you need a custom renderer, you may implement your own by extending `EntityRenderer`. Like all kinds of rendering, this code will be in your `client` source set.
```
its-miroma, line 107:
```suggestion
Like most entities, your projectile will also need a renderer to render its visuals while it's on the move. For this guide, we'll be using Minecraft's built-in `ThrownItemRenderer`.

::: tip

If you need a custom renderer, you may implement your own by extending `EntityRenderer`.

:::

Like all kinds of rendering, this code will be in your `client` source set.
```

uhmm, why does this section exist? Why not mention this below when you're registering?

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Registration {#registration}
+
+Now that we have a projectile entity and an item, we need to register them. In addition to these, we'll also register the entity renderer.
+
+For this example, our registration code will be in our normal and client initializers. You may organize your code as you see fit.
```
its-miroma:
```suggestion
For this example, our registration code will be in our main and client initializers.
```

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+
+Finally, we consume one item from the stack and mark the interaction as successful.
+
+<<< @/reference/latest/src/main/java/com/example/docs/projectile/HotTaterItem.java#use
+
+## Entity Renderer {#entity-renderer}
+
+Like most entities, your projectile will also need a renderer to render its visuals while it's on the move. For this guide, we'll be using Minecraft's built-in `ThrownItemRenderer`. If you need a custom renderer, you may implement your own by extending `EntityRenderer`. Like all kinds of rendering, this code will be in your `client` source set.
+
+## Registration {#registration}
```
its-miroma, line 109: Should we register things right after we create them instead? iirc other articles will say "Don't forget to register [thing] in your mod/client initializer"

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+
+**Client Initializer:**
+
+<<< @/reference/latest/src/client/java/com/example/docs/projectile/ExampleModProjectileClient.java#entrypoint
+
+Let's analyze the code.
+
+For convenience, we define a shared identifier for both the entity and the item.
+
+<<< @/reference/latest/src/main/java/com/example/docs/projectile/ExampleModProjectile.java#identifier
```
its-miroma: I would move this before the registration snippets.

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+
+**Initializer:**
+
+<<< @/reference/latest/src/main/java/com/example/docs/projectile/ExampleModProjectile.java#entrypoint
+
+**Client Initializer:**
+
+<<< @/reference/latest/src/client/java/com/example/docs/projectile/ExampleModProjectileClient.java#entrypoint
+
+Let's analyze the code.
```
its-miroma, line 127: You show the result first, then explain the significant parts in later sections..

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+### Registering the Item {#registering-the-item}
+
+We register the item using the `register` method from our `ModItems` class, created in the [Creating Your First Item](../items/first-item#preparing-your-items-class) guide. This code will also be in your `main` source set.
+
+<<< @/reference/latest/src/main/java/com/example/docs/projectile/ExampleModProjectile.java#register_item
+
+We also add the item to a creative tab for easy access.
+
+<<< @/reference/latest/src/main/java/com/example/docs/projectile/ExampleModProjectile.java#creative_tab
+
```
its-miroma, line 144: `#registering-the-entity` and `#registering-the-item` are redundant given the Creating Your First Item/Entity pages. These are things done to all items. This is also why I said to drop this separate section and replace it with "Don't forget to register [the item/the entity]:"

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+We use the Minecraft-provided `ThrownItemRenderer` for our entity. This is the same renderer used by throwable items like the snowball.
+
+<<< @/reference/latest/src/client/java/com/example/docs/projectile/ExampleModProjectileClient.java#renderer
+
+At this point, you can obtain the projectile item in game and test out the functionality.
+
+## Finalizing the Item {#finalizing-the-item}
+
+Even though our projectile works, it still doesn't have a model, texture, client item or a name. Items require a [model](../items/first-item#adding-a-model), [texture](../items/first-item#adding-a-texture), and [client item](../items/first-item#creating-the-client-item) with the name `hot_tater` in order to render correctly. You'll also want a [translation](../items/first-item#naming-the-item) to give the item a proper name. An example texture is provided below.
```
its-miroma:
```suggestion
## Finalizing the Item {#finalizing-the-item}

At this point, you can obtain the projectile item in game and test out the functionality.

However, you'll notice it still doesn't have a [model](../items/first-item#adding-a-model), [texture](../items/first-item#adding-a-texture), [client item](../items/first-item#creating-the-client-item) or a [name](../items/first-item#naming-the-item). Let's create the files using the identifier `hot_tater` in order to render everything properly. An example texture is provided below.
```

develop/entities/projectiles.md
```diff
@@ -0,0 +1,165 @@
...
+
+<DownloadEntry visualURL="/assets/develop/projectiles/hot_tater.png" downloadURL="/assets/develop/projectiles/hot_tater_small.png">Texture</DownloadEntry>
+
+Now you can test your projectile in game.
```
its-miroma:
```suggestion
Now you can test your projectile in-game:
```

its-miroma (changes requested):

develop/entities/projectiles.md
cassiancc: Shouldn't be anything, beyond a brief mention that data attachments exist.

dicedpixels: I removed the "(except Neo exclusives)" part.

develop/entities/projectiles.md
dicedpixels: We can but I really don't think that's the focus of this tutorial. Also yes, this is kind of a fireball, except you can't have it as an item.

develop/entities/projectiles.md
dicedpixels: I like a info box too. Switched.

develop/entities/projectiles.md
dicedpixels: Yep, 4 levels seemed excessive. Let me know which way is better.

develop/entities/projectiles.md
dicedpixels: Ah something I missed. I was originally planning to ignite the top. Fixed.

develop/entities/projectiles.md
dicedpixels: (See https://github.com/FabricMC/fabric-docs/pull/714/changes#r3880861277)

develop/entities/projectiles.md
dicedpixels: It will hit a block and start to sink vertically down. I'm not sure if we should mention this "sinking" behavior. I made an edit to make it not specific to entities.

develop/entities/projectiles.md
dicedpixels: No preference, I for one are confused by both :clueless:.

develop/entities/projectiles.md
dicedpixels: Wording was from ChampionAsh, but I assume air and fluids maybe.

develop/entities/projectiles.md
dicedpixels: (See https://github.com/FabricMC/fabric-docs/pull/714/changes#r3880889614)

develop/entities/projectiles.md
dicedpixels: (See https://github.com/FabricMC/fabric-docs/pull/714/changes#r3880889614)

develop/entities/projectiles.md
dicedpixels: (See https://github.com/FabricMC/fabric-docs/pull/714/changes#r3880889614)

develop/entities/projectiles.md
dicedpixels: Done.

develop/entities/projectiles.md
dicedpixels: (See https://github.com/FabricMC/fabric-docs/pull/714/changes#r3880889614)

develop/entities/projectiles.md
dicedpixels: I thought that would give the reader a general idea of what they're going to do. If they need further explanation, they can continue reading.

develop/entities/projectiles.md
dicedpixels: We can but there's kind of a circular dependency between the entity and the item. I felt like registering them at last would make the most sense.

develop/entities/projectiles.md
dicedpixels: If there's no objection, we can remove both and replace it with a small paragraph linking to the said pages.

reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java
dicedpixels: Leaving this for further discussion. I kept it so that people will notice it and not wonder why there are 3000 yellow lines.

reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java
dicedpixels: * First constructor is the one entity registration uses.
* The other two are convenience constructors for spawning, where the type is always the Hot Tater.
* Level being first matches vanilla's `Projectile.ProjectileFactory.create(ServerLevel, LivingEntity, ItemStack)`, which is what allows `HotTaterEntity::new` as a method ref in `spawnProjectileFromRotation`

reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java
dicedpixels: We could, I went ahead like this because it's there in `MiniGolemEntity#tick` too.

reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java
dicedpixels: Not sure if I understand the question but it should fail whenever the spot adjacent to the "hit" face already has something in it.

reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java
dicedpixels: Fixed.

reference/latest/src/main/java/com/example/docs/projectile/HotTaterItem.java
dicedpixels: Done.

sidebar_translations.json
dicedpixels: Lets go with "Custom Projectiles", at least it will match most of the titles.

its-miroma: Consider changing the page title too

reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java
its-miroma: yea, my bad, the question was unclear: if the first block that the projectile hits is `hitResult.getBlockPos()`, one could assume that every block between the start and the target, along `hitResult.getDirection()`, was empty, no?

If `level.isEmptyBlock(pos)` ever returned false, then the projectile would have hit that block before hitting the next one...

I don't think this makes my question clearer :sob:

reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java
its-miroma: But why did Checkstyle not complain?

develop/entities/projectiles.md
its-miroma: Out of scope - fair

develop/entities/projectiles.md
its-miroma: With my recommendation it wouldn't be 4 levels of headings, because I would remove the `### Overrides` grouping. (see: https://github.com/FabricMC/fabric-docs/pull/714#discussion_r3875479087)

develop/entities/projectiles.md
cassiancc: Sticking to naming conventions used in Minecraft itself makes sense to me.

reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java
cassiancc: That wouldn't be the case if it hits a block that does not have collision - I replicated this myself when throwing at Tall Grass blocks.

develop/entities/projectiles.md
its-miroma: of course, I'm saying in the definition: something like

```suggestion
- **Y-offset**: Pitch (rotation in the y direction).
```

reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java
its-miroma: Thank you, that makes sense!

develop/entities/projectiles.md
```diff
@@ -0,0 +1,161 @@
...
+
+It's a standard item implementation, with some special methods from `ProjectileItem`. Let's look at them:
+
+### Overriding `asProjectile()` {#override-as-projectile}
+
+This method converts the item into its entity form.
+
+<<< @/reference/latest/src/main/java/com/example/docs/projectile/HotTaterItem.java#as_projectile
+
+### Overriding `use()` {#pverride-use}
```
its-miroma:
```suggestion
### Overriding `use()` {#override-use}
```

its-miroma (changes requested):

reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java
its-miroma: That makes sense

reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java
dicedpixels: https://checkstyle.sourceforge.io/checks/coding/requirethis.html
Doesn't say anything about inherited methods. I think Checkstyle is not aware of the class hierarchy and just does file level analysis.

its-miroma: Even if it only did an analysis at the file level, I am not aware of any situation in which `this.` is wrong (excluding static shenanigans). I would have expected `;discard();` to be flagged...

cassiancc (dismissed):

reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java
```diff
@@ -0,0 +1,83 @@
...
+	protected Item getDefaultItem() {
+		return ExampleModProjectile.HOT_TATER_ITEM;
+	}
+	// #endregion default_item
+
+	// #region on_hit_block
+	@Override
+	protected void onHitBlock(BlockHitResult hitResult) {
+		super.onHitBlock(hitResult);
+		Level level = level();
```
its-miroma, line 51: I am still not sure why Checkstyle is not warning about `this.level()`...

its-miroma (changes requested):
The article LGTM as soon as my other unresolved threads are addressed.

ChampionAsh5357 also wanted to have a final look I believe.

develop/entities/projectiles.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Overriding `use()` {#override-use}
+
+Defines the action that happens when the item is used. In our case, we call the `Projectile.spawnProjectileFromRotation()` utility method to spawn the projectile.
+
+In addition to the standard parameters (level, item stack, and player), this utility method takes three additional floats:
+
+- **Y-Offset**: Rotation in the y direction (looking up or down).
+- **Power**: Scales how fast the projectile moves in blocks.
+- **Uncertainty**: Range of directions that the projectile can randomly go in when shot.
```
ekulxam: Calling this “range of directions” sounds weird to me. Perhaps we could reword this to include spread or a mention of decreased precision?

https://discord.com/channels/507304429255393322/1208846408552030238/1547545665863950388

ekulxam (changes requested):

develop/entities/projectiles.md
```diff
@@ -0,0 +1,161 @@
...
+
+<<< @/reference/latest/src/main/java/com/example/docs/projectile/HotTaterItem.java#as_projectile
+
+### Overriding `use()` {#override-use}
+
+Defines the action that happens when the item is used. In our case, we call the `Projectile.spawnProjectileFromRotation()` utility method to spawn the projectile.
+
+In addition to the standard parameters (level, item stack, and player), this utility method takes three additional floats:
+
+- **Y-Offset**: Rotation in the y direction (looking up or down).
```
ekulxam: I agree, there should be some mention of pitch. However, this parameter is used weirdly so it should probably be explained more clearly https://discord.com/channels/507304429255393322/1208846408552030238/1547550739910627358

ekulxam (changes requested):

develop/entities/projectiles.md
ekulxam: https://discord.com/channels/507304429255393322/1208846408552030238/1547554881903136870

its-miroma:
````suggestion
- `yOffset`: Offset for the pitch (rotation around the X axis, upward or downward), in degrees. Negative values angle the initial velocity upward.
- `pow`: Multiplier for the speed of the projectile movement.
- `uncertainty`: Imprecision of the projectile. 0 means no random spread. As the value increases, the thrown projectiles get more dispersed.

For more information, see the [Minecraft Wiki's article on projectiles](https://minecraft.wiki/w/Projectile#Initial_conditions).

::: details Why is the parameter called `yOffset`?

We don't know either, dear reader. Despite the name, the offset is applied to the pitch - to the velocity rotation around the X axis:

```java
      float yd = -Mth.sin((xRot + yOffset) * (float) (Math.PI / 180.0));
```

For example, vanilla uses a [`yOffset` of `-20.0F` for splash potions](https://mcsrc.dev/2/26.2/net/minecraft/world/item/ThrowablePotionItem#L28), which changes the initial velocity pitch from `source.getXRot()` to `source.getXRot() - 20.0F` (20 degrees upward, toward the sky).

Perhaps a more apt name would have been `xRotOffset`.

:::

````

Let's also add the Projectile article to resources in the frontmatter. And please add Earthcomputer and ekulxam to authors and Co-authors if you can.

**Review by its-miroma — CHANGES_REQUESTED** [review]

File: develop/entities/projectiles.md (hunk 44 above)
Comment (its-miroma) [reply develop/entities/projectiles.md]: @ekulxam Does this work for you?
[(later comment)](https://github.com/FabricMC/fabric-docs/pull/714#discussion_r3978726162)

File: develop/entities/projectiles.md (hunk 39 above)
Comment (dicedpixels) [reply develop/entities/projectiles.md:136]: @its-miroma Any action to be taken here?

File: develop/entities/projectiles.md (hunk 34 above)
Comment (dicedpixels) [reply develop/entities/projectiles.md]: Any further changes required here @its-miroma?

File: develop/entities/projectiles.md [hunk 46]
```diff
@@ -0,0 +1,179 @@
...
+
+### Overriding `use()` {#override-use}
+
+Defines the action that happens when the item is used. In our case, we call the `Projectile.spawnProjectileFromRotation()` utility method to spawn the projectile.
+
+In addition to the standard parameters (level, item stack, and player), this utility method takes three additional floats:
+
+- `yOffset`: Offset for the pitch (rotation around the X axis, upward or downward), in degrees. Negative values angle the initial velocity upward.
+- `pow`: Multiplier for the speed of the projectile movement.
+- `uncertainty`: Quantification of the inaccuracy (spread of directions) of the projectile. `0` means no random spread.
```
its-miroma:
```suggestion
- `uncertainty`: Imprecision of the projectile. 0 means no random spread. As the value increases, the thrown projectiles get more dispersed.
```

its-miroma (changes requested):

develop/entities/projectiles.md
its-miroma: Yes, the tip container

develop/entities/projectiles.md
```diff
@@ -0,0 +1,179 @@
...
+
+### Overriding `use()` {#override-use}
+
+Defines the action that happens when the item is used. In our case, we call the `Projectile.spawnProjectileFromRotation()` utility method to spawn the projectile.
+
+In addition to the standard parameters (level, item stack, and player), this utility method takes three additional floats:
+
+- `yOffset`: Offset for the pitch (rotation around the X axis, upward or downward), in degrees. Negative values angle the initial velocity upward.
+- `pow`: Multiplier for the speed of the projectile movement.
+- `uncertainty`: Imprecision of the projectile. 0 means no random spread. As the value increases, the thrown projectiles get more dispersed.
```
its-miroma:
```suggestion
- `uncertainty`: Imprecision of the projectile. 0 means no random spread. As the value increases, projectiles will be more dispersed, even when thrown with the same initial position and rotation.
```

develop/entities/projectiles.md
```diff
@@ -0,0 +1,179 @@
...
+
+<<< @/reference/latest/src/main/java/com/example/docs/projectile/HotTaterItem.java#use
+
+## Entity Renderer {#entity-renderer}
+
+Like most entities, your projectile will also need a renderer to render its visuals while it's on the move. For this guide, we'll be using Minecraft's built-in `ThrownItemRenderer`. If you need a custom renderer, you may implement your own by extending `EntityRenderer`. Like all kinds of rendering, this code will be in your `client` source set.
+
+## Registration {#registration}
+
+Now that we have a projectile entity and an item, we need to register them. In addition to these, we'll also register the entity renderer.
```
its-miroma:
```suggestion
Now that we have a projectile entity and an item, we need to register them as usual. In addition to these, we'll also register the entity renderer.
```

develop/entities/projectiles.md
```diff
@@ -0,0 +1,179 @@
...
+
+For this example, our registration code will be in our main and client initializers.
+
+For convenience, we define a shared identifier for both the entity and the item.
```
its-miroma:
```suggestion
For convenience, we define a shared identifier for both the entity and the item:
```

Are we not supposed to use one of the Ids classes? what is done for entities+items that share the same id, do we create `ModEntityItemIds` like `ModBlockItemIds`? :sob:

develop/entities/projectiles.md
```diff
@@ -0,0 +1,179 @@
...
+
+<<< @/reference/latest/src/client/java/com/example/docs/projectile/ExampleModProjectileClient.java#entrypoint
+
+Let's analyze the code.
+
+### Registering the Entity {#registering-the-entity}
+
+Like other entities, we use `EntityType.Builder`. We call the `of()` method with a hitbox size, client tracking range and tick update interval. Finally we build the entity type by passing a `ResourceKey`. This entity type is then registered to the `ENTITY_TYPE` registry. This code will be in your `main` source set.
+
+<<< @/reference/latest/src/main/java/com/example/docs/projectile/ExampleModProjectile.java#register_entity
```
its-miroma: Can you use the helpers in `ModEntityTypes`, like the first-entity article?

develop/entities/projectiles.md
its-miroma: The issue with those two sections + `#finalizing-the-item` is that they duplicate stuff from first-item and first-entity, and they do some things differently (see other comment). Imo, we replace `#registration` and `#finalizing-the-item` with a single short section that says:

Register the item, like we did [here]:
(show the code)

Don't forget to add [model], [texture], [client item], [name]:
(show the texture file)

Register the entity too, like we did [here]:
(show the code)

Let's use the vanilla `ThrownItemRenderer`:
(show the code)

And you're done!
(show the clip)

its-miroma (changes requested):

develop/entities/projectiles.md
dicedpixels: At a glace, it seems like `ModEntityTypes` wasn't designed to be a helper.`register` is private and `registerAttributes` is hardcoded for the Mini Golem.

its-miroma: hmm

its-miroma: hmm

its-miroma: Looking again, `ModEntityTypes` is where you're expected to register entity types (as static fields) and attributes (in `registerAttributes()`):

```java
	// in ModEntityTypes:
	public static final EntityType<HotTaterEntity> HOT_TATER = register(
			"hot_tater",
			EntityType.Builder
				.<HotTaterEntity>of(HotTaterEntity::new, MobCategory.MISC)
				.sized(0.25F, 0.25F)
				.clientTrackingRange(4)
				.updateInterval(10)
	);
```

It's done so you don't call `ResourceKey#create` and `Registry#register`.

Maybe it should be refactored to a helper that exposes `registerEntityType(...)`, `registerAttributes(...)`, `registerRenderer(...)`, tho out of scope

reference/latest/src/main/java/com/example/docs/projectile/HotTaterItem.java
```diff
@@ -0,0 +1,57 @@
...
+		level.playSound(null, player.getX(), player.getY(), player.getZ(), SoundEvents.SNOWBALL_THROW, SoundSource.NEUTRAL, 0.5F, 0.4F / (level.getRandom().nextFloat() * 0.4F + 0.8F));
+
+		// Spawn the projectile on the server only.
+		if (level instanceof ServerLevel serverLevel) {
+			Projectile.spawnProjectileFromRotation(HotTaterEntity::new,
+					serverLevel,
+					itemStack,
+					player,
+					/* yOffset: */ 0.0F,
+					/* power: */ 1.5F,
```
its-miroma:
```suggestion
					/* pow: */ 1.5F,
```

reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java
its-miroma, line 54: ~~Might be a dumb question: can we use `level instanceof ServerLevel serverLevel` here?~~ edit: no

develop/entities/projectiles.md
```diff
@@ -0,0 +1,149 @@
...
+
+<<< @/reference/latest/src/main/java/com/example/docs/projectile/ExampleModProjectile.java#identifier
+
+<<< @/reference/latest/src/main/java/com/example/docs/projectile/ExampleModProjectile.java#register_item
+
+We also add the item to a creative tab for easy access, like we did in the [Adding the Item to a Creative Tab](../items/first-item#adding-the-item-to-a-creative-tab) guide:
+
+<<< @/reference/latest/src/main/java/com/example/docs/projectile/ExampleModProjectile.java#creative_tab
+
+Don't forget to add a [model](../items/first-item#adding-a-model), [texture](../items/first-item#adding-a-texture), [client item](../items/first-item#creating-the-client-item), and [name](../items/first-item#naming-the-item), using the identifier `hot_tater`:
```
its-miroma:
```suggestion
Don't forget to add a [model](../items/first-item#adding-a-model), [texture](../items/first-item#adding-a-texture), [client item](../items/first-item#creating-the-client-item), and [name](../items/first-item#naming-the-item), using the identifier `hot_tater`. You should also [add the item to a creative tab](../items/first-item#adding-the-item-to-a-creative-tab). Here's an example texture:
```

What do you think?

reference/latest/src/main/java/com/example/docs/projectile/ExampleModProjectile.java
```diff
@@ -0,0 +1,52 @@
...
+					new Item.Properties().stacksTo(16));
+	// #endregion register_item
+
+	@Override
+	public void onInitialize() {
+		// #region creative_tab
+		CreativeModeTabEvents.modifyOutputEvent(ModItems.CUSTOM_CREATIVE_TAB_KEY).register(tab -> {
+			tab.accept(HOT_TATER_ITEM);
+		});
+		// #endregion creative_tab
```
its-miroma:
```suggestion
		CreativeModeTabEvents.modifyOutputEvent(ModItems.CUSTOM_CREATIVE_TAB_KEY).register(tab -> {
			tab.accept(HOT_TATER_ITEM);
		});
```

public/assets/develop/projectiles/hot_tater.png
its-miroma, line 1: Can you rename this to `hot_tater_preview` or sth, and name the downloadable texture `hot_tater`? This way when the user downloads it they get a file with the correct name. Please also update the DownloadEntry accordingly.

Related: #707

develop/entities/projectiles.md
```diff
@@ -0,0 +1,149 @@
...
+
+<<< @/reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java#constructors
+
+### Overriding `getDefaultItem()` {#override-get-default-item}
+
+Defines the item form of this projectile.
+
+<<< @/reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java#default_item
+
+::: info
```
its-miroma:
```suggestion
::: warning IMPORTANT
```

blends in too much otherwise

develop/entities/projectiles.md
```diff
@@ -0,0 +1,149 @@
...
+
+- `yOffset`: Offset for the pitch (rotation around the X axis, upward or downward), in degrees. Negative values angle the initial velocity upward.
+- `pow`: Multiplier for the speed of the projectile movement.
+- `uncertainty`: Imprecision of the projectile. 0 means no random spread. As the value increases, projectiles will be more dispersed, even when thrown with the same initial position and rotation.
+
+For more information, see the [Minecraft Wiki's article on projectiles](https://minecraft.wiki/w/Projectile#Initial_conditions).
+
+::: details Why is the parameter called `yOffset`?
+
+We don't know either, dear reader. Despite the name, the offset is applied to the pitch, the rotation of the velocity around the X axis:
```
its-miroma:
```suggestion
We don't know either, dear reader. Despite the name, [the offset is applied to the pitch](https://mcsrc.dev/2/26.2/net/minecraft/world/entity/projectile/Projectile#L156), which is the rotation of the *velocity* around the X axis:
```

develop/entities/projectiles.md
```diff
@@ -0,0 +1,149 @@
...
+
+```java
+float yd = -Mth.sin((xRot + yOffset) * (float) (Math.PI / 180.0));
+```
+
+For example, vanilla uses a [`yOffset` of `-20.0F` for splash potions](https://mcsrc.dev/2/26.2/net/minecraft/world/item/ThrowablePotionItem#L28), which changes the initial velocity pitch from `source.getXRot()` to `source.getXRot() - 20.0F` (20 degrees upward, toward the sky).
```
its-miroma:
```suggestion
For example, when vanilla uses a [`yOffset` of `-20.0F` for splash potions](https://mcsrc.dev/2/26.2/net/minecraft/world/item/ThrowablePotionItem#L28), it changes the initial velocity pitch from `source.getXRot()` to `source.getXRot() - 20.0F` (20 degrees upward, toward the sky).
```

its-miroma (changes requested):

develop/entities/projectiles.md
dicedpixels: We could refactor the entity code the next time there's an entity to be added. I can create an issue if you want.

its-miroma: that seems sensible :+1:

we should still register the entity as a static field in ModEntityTypes tho

public/assets/develop/projectiles/hot_tater.png
dicedpixels: Done!

reference/latest/src/main/java/com/example/docs/projectile/HotTaterEntity.java
dicedpixels: As discussed in the Discord server, they're not equal and we should stick to the current way to check the side.

reference/latest/src/main/java/com/example/docs/projectile/ExampleModProjectile.java
```diff
@@ -0,0 +1,35 @@
...
+import net.fabricmc.fabric.api.creativetab.v1.CreativeModeTabEvents;
+
+import com.example.docs.ExampleMod;
+import com.example.docs.item.ModItems;
+
+// #region entrypoint
+public class ExampleModProjectile implements ModInitializer {
+	// #region identifier
+	// Identifier for the item. The entity type uses the same path via ModEntityTypes.
+	public static final Identifier HOT_TATER_ID = Identifier.fromNamespaceAndPath(ExampleMod.MOD_ID, "hot_tater");
```
cassiancc: This should be moved to `ModItemIds` now that it is no longer used for entities.

cassiancc (changes requested):

reference/latest/src/main/java/com/example/docs/projectile/ExampleModProjectile.java
its-miroma: should we also call the item `HOT_TATER` instead of `HOT_TATER_ITEM`?

In general, what is the protocol for using the same identifier for an entity and an item, etc? I've only seen block+item and in that case we have a whole separate class...

Because rn there's a stray `"hot_tater"` in the call to `ModEntityTypes#register`

its-miroma (approved):
Could you also open that issue about `ModEntityTypeIds`? And also about `ModEntityTypes` that we said might be weird

cassiancc (approved):

PR #736, opened 2026-09-12
Fix lightning spawning on the networking page

### LINKED ISSUE #706: Networking Tutorial Lightning Spawning broken
In 26.2, you now need to change the lightning entities id on the client before spawning it to prevent an error that is thrown when getting the id of an entity when it's 0 (which it always is on the client). I think spawning an entity on the client is not a good example anyway as it is not a common use case for networking (I don't have a better idea right now though)
See https://discord.com/channels/507304429255393322/507982478276034570/1538285477830991882

skippyall:
Closes #706 
Adds an id to the packet and uses it to spawn the lightning on the client. The text was updated for this change.


develop/networking.md
```diff
@@ -142,8 +142,8 @@ lambda expression).
 
 Let's examine the code above.
 
-We can access the data from our payload by calling the Record's getter methods. In this case `payload.pos()`. Which then
-can be used to get the `x`, `y` and `z` positions.
+We can access the data from our payload by calling the Record's getter methods. In this case `payload.pos()`, which then
+can be used to get the `x`, `y` and `z` positions, and `payload.id()` for the entity id.
```
its-miroma:
```suggestion
We can access the data from our payload by calling the Record's getter methods. In this case, `payload.pos()`
can be used to get the `x`, `y` and `z` positions, and `payload.id()` for the entity ID:
```

develop/networking.md
```diff
@@ -106,7 +106,7 @@ We only send packets when the action is initiated on the server, by returning ea
 
 <<< @/reference/latest/src/main/java/com/example/docs/networking/basic/LightningTaterItem.java#client_check
 
-We create an instance of the payload with the user's position:
+We create an instance of the payload with the user's position and the entity id for the lightning:
```
its-miroma:
```suggestion
We create an instance of the payload with the user's position and the entity ID for the lightning:
```

its-miroma (changes requested):

its-miroma (approved):
We can merge this so the example doesn't crash, but I'll keep the issue open because we need to come up with a better example.

cassiancc (approved):

PR #738, opened 2026-09-15
Port to 26.3

cassiancc:
Here we go again! I'd recommend focusing on individual commits to avoid the auto-generated moving around.
- [x] Mod compiles
- [x] Decide whether to show item component modification API or registering compostables inline.
- [x] Fix data generation being unable to run due to missing loot table
- [x] 26.3 porting docs
- [x] Potion recipe providers (ideally https://github.com/FabricMC/fabric-api/pull/5595 is merged before the Docs)
- [ ] Determine wording on ~~configured~~ feature article.
- [ ] Advancement docs should clarify display/rootDisplay
- [ ] Data generation docs for int providers? 
- [ ] We're out of memory again!


reference/latest/src/client/java/com/example/docs/keymapping/ExampleModKeyMappingsClient.java
```diff
@@ -27,7 +27,7 @@ public class ExampleModKeyMappingsClient implements ClientModInitializer {
 	KeyMapping sendToChatKey = KeyMappingHelper.registerKeyMapping(
 		new KeyMapping(
 				"key.example-mod.send_to_chat", // The translation key for the key mapping.
-				InputConstants.Type.KEYSYM, // The type of the keybinding; KEYSYM for keyboard, MOUSE for mouse.
+				InputConstants.Type.KEYBOARD, // The type of the keybinding; KEYSYM for keyboard, MOUSE for mouse.
```
its-miroma:
```suggestion
				InputConstants.Type.KEYBOARD, // The type of the keybinding; KEYBOARD for keyboard, MOUSE for mouse.
```

reference/latest/src/main/java/com/example/docs/item/ModItems.java
```diff
@@ -250,7 +253,7 @@ public class ModItems {
 	// #region axe
 	public static final Item GUIDITE_AXE = register(
 					ModItemIds.GUIDITE_AXE,
-					settings -> new AxeItem(GUIDITE_TOOL_MATERIAL, 5.0F, -3.0F, settings),
+					settings -> new Item(settings.axe(GUIDITE_TOOL_MATERIAL, 5.0F, -3.0F)),
```
its-miroma, line 256: The page probably needs to be updated

reference/latest/src/main/resources/fabric.mod.json
```diff
@@ -60,7 +60,7 @@
   ],
   "depends": {
     "fabricloader": ">=0.19.0",
-    "minecraft": "~26.2",
+    "minecraft": ">26.2",
```
its-miroma: I'll update bump.ts to also change this. Why not use `~26.3` or similar? I also assume it cannot reference the value in build.gradle, right?

reference/latest/src/client/java/com/example/docs/keymapping/ExampleModKeyMappingsClient.java
CelDaemon: I would omit the KEYBOARD reference entirely. Unlike KEYSYM, KEYBOARD is already pretty clear on its own.

reference/latest/src/main/resources/fabric.mod.json
cassiancc: Solely because I started the port during the snapshot phase, and `~26.3` excludes snapshots. I believe it can, though I usually don't.

reference/latest/src/client/java/com/example/docs/datagen/ExampleModBlockLootTableProvider.java
```diff
@@ -33,7 +32,7 @@ public void generate() {
 		dropWhenSilkTouch(ModBlocks.PRISMARINE_LAMP);
 		// Make condensed oak logs drop between 7 and 9 oak logs
 		add(ModBlocks.CONDENSED_OAK_LOG, LootTable.lootTable().withPool(applyExplosionCondition(Items.OAK_LOG, LootPool.lootPool()
-				.setRolls(new UniformGenerator(new ConstantValue(7), new ConstantValue(9)))
+				.setRolls(ContextIntProviders.avg(ContextIntProviders.exactly(7), ContextIntProviders.exactly(9)))
```
Earthcomputer:
```suggestion
				.setRolls(ContextIntProviders.between(ContextIntProviders.exactly(7), ContextIntProviders.exactly(9)))
```

reference/latest/src/main/java/com/example/docs/worldgen/ExampleModWorldPlacedFeatures.java
```diff
@@ -41,7 +41,7 @@ public class ExampleModWorldPlacedFeatures {
 	public static void configure(BootstrapContext<PlacedFeature> context) {
 		// #endregion datagen_world_placed_features_class
 		// #region datagen_world_conf_feature_register
-		HolderGetter<ConfiguredFeature<?, ?>> configuredFeatures = context.lookup(Registries.CONFIGURED_FEATURE);
+		HolderGetter<Feature> configuredFeatures = context.lookup(Registries.FEATURE);
```
Earthcomputer:
```suggestion
		HolderGetter<Feature> features = context.lookup(Registries.FEATURE);
```

the region name could maybe be changed too although that's less important

Earthcomputer (changes requested):

